Issue #3, opened 2016-10-20
Let programs decide the log level/config (and not pykeepass)


harlowja:
Seems like there is a stray 'logging.basicConfig(level=logging.DEBUG)' when you import pykeepass; this is sorta a pain in the ass since my application using pykeepass wants to setup the logging level at program execution time, and not have pykeepass try to set it up itself...

Can we please remove that line & let applications using this library do the configuration (when they want). Thanks.

bronger:
Thanks to @Evidlo, it re-appeared the second time. See PR https://github.com/libkeepass/pykeepass/pull/191.

PR #5, opened 2016-11-10
Remove main library import logging configuration/setup

harlowja:
This seems to have reappeared!

So kill it again ;)

See issue #3


pschmitt:
Oh yeah, thanks.

PR #9, opened 2017-03-09
Bugfix + New Stuff

Evidlo:
`entries` throws an error when the database is empty because `res` is `None` and is not iterable.

```
Open file /home/evan/.passhole.kdbx
  File "/home/evan/bin//passhole", line 77, in <module>
    dmenu(args.dmenu, args.tabbed)
  File "/home/evan/bin//passhole", line 48, in dmenu
    print(kp.entries)
  File "/home/evan/resources/venv/lib/python2.7/site-packages/pykeepass/pykeepass.py", line 52, in entries
    return [x for x in res if not x.is_a_history_entry]
TypeError: 'NoneType' object is not iterable
```


Evidlo:
You should consider making `entries`, `groups`, and all the `find_` functions return an empty list instead of None so that the user doesn't have to explicitly check the return value when using these functions.

Evidlo:
I spent a couple hours rewriting a few things and adding tests.  Below is a summary:

- added `delete` function to `Entry` and `Group`
- added tests
- deleted all `find_entry_by_*` and `find_group_by_*` and replaced with a `first` argument to `find_entries_by_*` and `find_groups_by_*`
  - when `first` is `True`, the return type is either an instance of the object or `None`
  - when `first` is `False` (the default), the return type is always a list
- deleted `get_root_group` and moved into `root_group` property
- changed all icon constants to strings

If there's anything you don't like or want me to change, just say the word.

Evidlo:
Have you had a chance to look over this?

pschmitt:
Yep, I just did. It's looking good to me, the only thing that is missing is the test database used for testing. Great work, thanks !

PR #10, opened 2017-03-20
Added missing test.kdbx and readme examples

Evidlo:
Changes I made:

- added missing test database
- added some api documentation + more examples to the readme
- removed extraneous `re` import
- the first argument of `add_group` and `add_entry` is expected to be a `Group` instance, not a string
  - I felt that this helps with improving api consistency


pschmitt:
Thanks!

PR #15, opened 2017-03-27
Set expires parameter to False by default

drcursor:
In order to bypass https://github.com/pschmitt/pykeepass/issues/14 we
set expires to False. This allows faulty versions of KeePassX to open
files with passwords added by pykeepass.


pschmitt:
I'm not a huge fan of implementing workarounds here because of faulty clients but since this one is trivial...

PR #17, opened 2017-04-03
Added set_password

Evidlo:
This adds support for setting the master password on the database.

This was taken from [add_credentials](https://github.com/libkeepass/libkeepass/blob/master/libkeepass/common.py#L178-L182) in libkeepass.

There is a potential issue, though.  It looks like libkeepass stores the hashed password (if it exists) and the hashed keyfile (if it exists) in kdb.keys.  If there's only one item in this list, I think it might be impossible to tell if it's a password or a keyfile hash.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -190,3 +190,16 @@ Adding Groups
    # save the database
    >>> kp.save()
        
+Miscellaneous
+-------------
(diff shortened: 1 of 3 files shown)


pschmitt:
Please rebase this against the current master branch.

PR #23, opened 2017-05-26
Python 3 print compatibility

eduarias:
On PR that fixed issue 20 the compatibility with print statement in Pyhton 3 was broken, when trying to print TypeError was raised:
`TypeError: __repr__ returned non-string (type bytes)`

I added unit test for both issue 20 and print statement, fix the issues and run them in python 2.7 and 3.5.


eduarias:
The PR #24 has also this changes and improves the testing

PR #29, opened 2017-06-08
added icon option when creating a group

### LINKED ISSUE #28: set explicitely the icon when creating a new group
It's currently not possible to set a different icon from the default one(18) when creating a new group.

mattLLVW:
This resolve #28 
I added the ability to specify an icon while creating a group.
The Group object now has the attribute icon.


pschmitt:
Thanks !

PR #30, opened 2017-06-10
exception when adding entry, resolve #27

### LINKED ISSUE #27: Entries are defined only by their group and by their title
Hello,
Currently, entries are only defined during creation by their tree(group) and title.
Even if their username and password is different it overwrite it.
With the force_creation, t's possible to, but i think it's safer to check also at least the username before overwriting it.
Would you accept a pull request on this?
Thank you

mattLLVW:
I added a new function that search for an entry based on its path, title and username. When trying to add an entry with the same attributes, it raise an exception unless force_creation is True.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -164,7 +164,7 @@ Adding Entries
 
 Adding Groups
 --------------
-**add_group** (destination_group, group_name)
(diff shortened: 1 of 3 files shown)


pschmitt:
lgtm

PR #32, opened 2017-06-17
Fixes #31

### LINKED ISSUE #31: Missing test_keyfile.kdbx for unit tests
Hi,
It seems that unit test are trying to use test_keyfile.kdbx, but it's not present in the tests directory.
As a result, test using it are not passing.

Evidlo:
Fixes issue #31 
- renamed test keyfile
- all tests now use both password and keyfile when opening test database
- removed print statements


pschmitt:
Thanks.

PR #34, opened 2017-06-25
Fixes #33

### LINKED ISSUE #33: add_entry function does not expose expiration settings
[@wget42](https://twitter.com/wget42) reached out to me on Twitter regarding this issue.
Here's a quick transcription:

- The [add_entry](https://github.com/pschmitt/pykeepass/blob/master/pykeepass/pykeepass.py#L256) function does not allow to set an expiration date

Evidlo:
- fixed #33 
- refactored some date stuff to use dateutil ~~builtin~~ (more robust than parsing with strptime)
- added setters for atime, ctime, mtime, expiry_time

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -43,8 +43,6 @@ Simple Example
 Finding Entries
 ----------------------
 
-The supported find commands are listed below
-
 **find_entries_by_title** (title, regex=False, tree=None, history=False, first=False)
 
 **find_entries_by_username** (username, regex=False, tree=None, history=False, first=False)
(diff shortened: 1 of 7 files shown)


Evidlo:
I think Travis is misconfigured.

Evidlo:
Initially I had assumed that dateutil shipped with python, but this is actually not the case.  However, I still think that letting dateutil do the datetime parsing will be more robust as the Keepass format allows for any date format, not just the one originally defined in `xmlfactory.py`

pschmitt:
The Travis CI config is fine. You need to add new dependencies to `requirements.txt`. And the package name should be `python-dateutil` not `dateutil`.

BTW could you please refrain from incrementing the version? I like to bump versions separately (and tag them in git).

pschmitt:
Thanks !

PR #37, opened 2017-07-10
Add flags for regex searching

### LINKED ISSUE #36: Case insensitive regex
This line:
lstEntry = kp.find_entries_by_notes(".*Chocolate.*",regex=True)
...finds the entry with notes:
Chocolate chips are a great invention

However, I cannot make the regex case-insensitive using the "i" modifier. I have tried these:
lstEntry = kp.find_entries_by_notes("/.*chocolate.*/i",regex=True)
lstEntry = kp.find_entries_by_notes("(.*chocolate.*)i",regex=True)

but both fail to find "Chocolate". Is this a bug in the regex or is there another way to make the search case insensitive?
Thanks

Evidlo:
Fixes #36

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -43,21 +43,23 @@ Simple Example
 Finding Entries
 ----------------------
 
-**find_entries_by_title** (title, regex=False, tree=None, history=False, first=False)
+**find_entries_by_title** (title, regex=False, flags=None, tree=None, history=False, first=False)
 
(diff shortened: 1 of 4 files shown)


pschmitt:
Nice, nice. Thanks !

Evidlo:
I messed up the two `optional flags` hyperlinks in README.md.  They should just be `flags`

PR #42, opened 2017-07-28
Fix gibberish passwords when saving a second time

wget:
This patch solves the problem described in the title (and described in the code as well).

As soon [this libkeepass PR is merged](https://github.com/libkeepass/libkeepass/pull/5), we will be able to go a bit further by relying only on the file descriptor opened by the `libkeepass` library. That way, we won't need to open a file descriptor by ourselves.

### DIFF
--- a/pykeepass/pykeepass.py
+++ b/pykeepass/pykeepass.py
@@ -39,13 +39,14 @@ class PyKeePass(object):
         ).__enter__()
 
     def save(self, filename=None):
-        # FIXME The *second* save operations creates gibberish passwords
-        # FIXME the save operation should be moved to libkeepass at some point
-        #       we shouldn't need to open another fd here just to write
         if not filename:
             filename = self.kdb_filename
         with open(filename, 'wb+') as outfile:
+            # fix issue 43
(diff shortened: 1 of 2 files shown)


Evidlo:
Could we move the discussion out of the code comments and into an issue?

wget:
@Evidlo Done. cf https://github.com/pschmitt/pykeepass/issues/43

pschmitt:
LGTM

wget:
@pschmitt Test added. I didn't know if you liked to have the test in the already available `add_entry` test or if you wanted me to create a dedicated one. If this doesn't suit your needs, let me know :)

wget:
It indeed introduced a regression. Gonna fix that.

pschmitt:
I'd rather have a dedicated test for this one.

wget:
@pschmitt Ok will do. But according to further testing I made,  I think this would be better to put the unprotecting part in `libkeepass` rather than `pykeepass`. Indeed the way field are protected differs between kdbx3 et kdbx4.

pylipp:
What's the status on this PR?

Evidlo:
This should be closed.

PR #47, opened 2017-07-30
lxml Element have no Value field

wget:
Creating an `Entry` manually then changing any of its field was failing with the following error message:

```
    results[0].Value = value
AttributeError: 'lxml.etree._Element' object has no attribute 'Value'
```

(only tested on Python 3 but I assume travis will do the job if it has tests for this use case).


pschmitt:
I don't believe there any tests for this. Would you mind implementing one?

pschmitt:
Obsoleted by #57

PR #49, opened 2017-08-02
Fix ctime setter

rubenv:


pschmitt:
Oh. Thanks !

Evidlo:
That's one my fault.  I never added tests for those setters

PR #55, opened 2017-10-26
resolve #54 implementation

### LINKED ISSUE #54: find entries by kwargs
Instead of having 8 different functions to find entries, it would be nice to just pass kwargs to `_find_entry_by`.

This way, you can actually find entries by multiple parameters, i.e:
`find_entries(title=exemple, username=me...)`

Would you consider a PR for this?

mattLLVW:
This PR expose find_entry_by ( #54 ), accept multiple kwargs and is retro-compatible.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -23,7 +23,7 @@ Simple Example
    >>> entry = kp.find_entries_by_title('facebook', first=True)
 
    # find entry by multiple arguments
-   >>> entry = kp.find_entries_by(Title='gmail', UserName='myusername', Password='myPassw0rdXX', first=True)
+   >>> entry = kp.find_entries_by(title='gmail', username='myusername', password='myPassw0rdXX', first=True)
 
    # retrieve the associated password
    >>> entry.password
@@ -77,7 +77,7 @@ This returns one entry except if ``history`` is set.
 
 **find_entry_by** (regex=False, flags=None, tree=None, history=False, first=False, \**kwargs)
 
-where \**kwargs accept multiple keyword arguments in ``Title``, ``UserName``, ``Password``, ``URL``, ``Notes``.
+where \**kwargs accept multiple keyword arguments in ``title``, ``username``, ``password``, ``url``, ``notes``.
 
 **entries**
 
--- a/README.rst
+++ b/README.rst
@@ -22,6 +22,9 @@ Simple Example
    # find any entry by its title
    >>> entry = kp.find_entries_by_title('facebook', first=True)
 
+   # find entry by multiple arguments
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -22,6 +22,9 @@ Simple Example
    # find any entry by its title
    >>> entry = kp.find_entries_by_title('facebook', first=True)
 
+   # find entry by multiple arguments
+   >>> entry = kp.find_entries_by(Title='gmail', UserName='myusername', Password='myPassw0rdXX', first=True)
```
pschmitt: I'd be better IMHO to lowercase the kwargs here. This means we need to maintain a dict that translates between kwargs and KeePass XML property names, but makes usage so much easier.

pykeepass/pykeepass.py
```diff
@@ -153,16 +153,26 @@ def delete_group(self, group):
...
-            xp = './/Entry/String/Key[text()="{}"]/../Value[re:match(text(), "{}", "{}")]/../..'.format(
-                key, value, flags
-            )
-        else:
-            xp = './/Entry/String/Key[text()="{}"]/../Value[text()="{}"]/../..'.format(
-                key, value
-            )
+    def find_entry_by(self, regex=False, flags=None,
+                       tree=None, history=False, first=False, **kwargs):
+        key_list = ['Title', 'UserName', 'Password', 'URL', 'Notes']
```
pschmitt: See previous comment.
```
keys = {'title': 'Title', 'username': 'UserName', 'password': 'Password'...}
```

mattLLVW:
Changes done

Evidlo:
Just thinking out loud here, but what if instead of `keys_dict`, we used something like this:

```
keys_xp = {
'title': '/String/Key[text()="Title"]/../Value[text()="{value}"]/../..',
'username': '/String/Key[text()="Title"]/../Value[text()="{value}"]/../..',
'password': '/String/Key[text()="Title"]/../Value[text()="{value}"]/../..',
'url': '/String/Key[text()="Title"]/../Value[text()="{value}"]/../..',
'notes': '/String/Key[text()="Title"]/../Value[text()="{value}"]/../..'
}
```

It's more verbose, but it gives us flexibility down the road to easily add support for searching other attributes like so:

```
keys_xp = {
    ...,
    ...,
    'autotype': 'Autotype/Enabled[text()="{value}"]',
    'creation_time': 'Times/CreationTime[text()="{value}"]'
    ...,
```

pschmitt:
@Evidlo Interesting idea.

pschmitt:
Obsoleted by #59

PR #57, opened 2017-10-26
Fix editing tags / icon (setting subelement text)

### LINKED ISSUE #56: Can't modify or create entry tags
`entry.tags = ['some', 'thing']` doesn't work. Tags stay the same. I think the error is in the `BaseElement._set_subelement_text` tried to fix it but no luck.

gpchelkin:
Remove subelement and create it with passed value. Fix #56

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -228,7 +228,7 @@ class EntryTests(unittest.TestCase):
         entry.url = changed_string + 'url'
         entry.notes = changed_string + 'notes'
 #        entry.expires = False
-        entry.expiry_time = changed_time
+#        entry.expiry_time = changed_time
         entry.icon = icons.GLOBE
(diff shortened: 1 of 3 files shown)


Evidlo:
Could you add a test to EntryTests for this? Maybe `test_set_fields`.  We should probably test setting all attributes there.

*edit* moved comment to PR

tests/tests.py
```diff
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime,
```
pschmitt: `datetime,` ?

gpchelkin: already fixed. yeah, I am watching travis logs too and trying to pass 😄

pschmitt:
LGTM

PR #59, opened 2017-10-28
Unify all find functions using xpath

Evidlo:
This is a prospective PR implementing what I talked about in #55.  We can now easily extend searching for more attributes just by adding a new line to `keys_xp` in `find_entries` or `find_groups`.  I've held off on adding stuff to the readme and more extensive tests until we sort out any issues here.

This PR unifies all `find_x_by_y` functions and adds `find_entries` and `find_groups` so that multiple parameters can be searched simultaneously.  For example

    kp.find_entries(path='foobar_group/', title='foo.*', url="http://example.com", regex=True)

returns all entries that begin with `foo` and match the URL inside `foobar_group`.

This necessarily introduces some breaking changes:
- renamed `find_entry_by_uuid` to `find_entries_by_uuid`
UUIDs are now searchable by regex and can return multiple entries. @laerus 
- deleted `_find_exact_entry` 
This behavior is replicated with `find_entries(name=, username=, tree=, first=True)` @mattLLVW 
- group paths *must* end with `/`
This is because the behavior of `find_entries_by_path('foo/bar')` is ambiguous.  Are we searching for entries in the group `bar`, or are we getting the entry called `bar` inside group `foo`?

On an unrelated note, should the repo description still say 'low level'?  I think this is the highest level Python library available for KeePass.


pykeepass/pykeepass.py
```diff
@@ -15,6 +15,7 @@
 from pykeepass.entry import Entry
 from pykeepass.group import Group
 
+logging.basicConfig(level=logging.INFO, format='%(message)s')
```
pschmitt: Don't !

pschmitt:
@Evidlo You are right. This isn't low-level anymore.

pschmitt:
Thanks

Evidlo:
I squashed all those extra commits together.

pschmitt:
bitmoji

PR #64, opened 2018-01-25
fix #63

### LINKED ISSUE #63: Groups order sometimes reversed
I have several subgroups.  And subgroups within subgroups.

When printing groups for example kp.groups.   Sometimes groups are in reverse order.  For example:
b'Group: "Bob/Internet"', b'Group: "Internet/Bob/Shopping"'

Is this on purpose?

Evidlo:
Fixes #63 

Path generating function got the append backwards.  Looks like it was never caught because we've never had groups or entries more than 2 deep in our tests.


pschmitt:
Neat. Thanks!

Issue #65, opened 2018-02-01
feature request: ability to search custom fields


leupo:
First of all: Thanks a lot for that lib.
My database entries are clearly idetified by a custom filed + username.  I can extract that custom fields with: get_custom_property() , but i want for search it. Could you implement that?

Evidlo:
This should be closed.

leupo:
Thank you so much Evidlo!

Evidlo:
@pschmitt Could you push this to pypi?

pschmitt:
Done: https://pypi.python.org/pypi/pykeepass/2.8.1

PR #67, opened 2018-02-04
add custom string field searching

Evidlo:
Implements #65


Evidlo:
Should I have called it `custom` instead of `string`?  The fact that `string` is a dict might be confusing.

PR #68, opened 2018-03-05
Add support for notes in add_group

wpoely86:
You can also set notes for groups. This PR adds support for that.


pschmitt:
Thanks, this is looking good. Only thing missing is a test case.

wpoely86:
@pschmitt I added a test in the group test. Is that sufficient?

pschmitt:
Thanks, but what's up with 8cce5d6?

PR #82, opened 2018-05-30
Rename tree

Evidlo:
Renames `tree` to `group` and adds documentation on searching groups.

Right now if people want to search a specific group, they have to dip down into LXML with `group._element`.  Now they can just provide the group directly.  I've also added docs on the new `recursive` argument.


pschmitt:
LGTM. Thanks.

Issue #44, opened 2017-07-28
Reuse the libkeepass context manager


wget:
Like [described in the code](https://github.com/pschmitt/pykeepass/blob/ee710b9b654aa40e3ed15456419c361043456edd/pykeepass/pykeepass.py#L43), pykeepass has to recreate a write file stream in order to write to the context manager, because the libkeepass [does only provide a context manager managing a read-only stream](https://github.com/libkeepass/libkeepass/pull/5).

Also, since PyKeePass has been developed in a declarative way (a la C) rather than the Pythonic way, the libkeepass context manager isn't reused and pykeepass fakes the use of the libkeypass context manager by calling its underlying method (`open.__enter__()`) manually.

And finally, when calling the resource manager manually, depending on the garbage collector implementation (e.g.: CPython), the file descriptor might not be freed correctly leading in resources/memory leaks. The garbage collector never calls `__exit__()` explicitly, only with-statements do that. It's generally bad to depend on the garbage collector to trigger cleanup code anyways. This is why closing the file stream from the underlying library is recommended (especially and the mechanisms included in that lib are being deconstructed by the piece of software making use of it).

wget:
After further investigation, it appears the `__del__` function is only called [after all references to the object have been deleted](https://stackoverflow.com/a/1481512/3514658).

In this use case calling the `close` method from the underlying libkeepass doesn't work as there is no reference to it any more. It this for this particular use case, Python is smart enough to figure out to close the stream used by the underlying library.

```
The password you provided is incorrect or the KeePass file is invalid or has been corrupted.
Exception ignored in: <bound method PyKeePass.__del__ of <pykeepass.pykeepass.PyKeePass object at 0x7fa9e357dbe0>>
Traceback (most recent call last):
  File "/home/wget/IT/Dev/batchs/python_libo_to_keepass/pykeepass/pykeepass.py", line 33, in __del__
    self.kdb.close()
AttributeError: 'PyKeePass' object has no attribute 'kdb'
```

Evidlo:
This should be closed.

PR #84, opened 2018-06-22
KDBX4 Support and other stuff

### LINKED ISSUE #43: Second time save creates gibberish passwords
Like [described in the code itself](https://github.com/pschmitt/pykeepass/blob/ee710b9b654aa40e3ed15456419c361043456edd/pykeepass/pykeepass.py#L42), if we have already saved the keepass database, saving yet another time will corrupt passwords.

This is due to the fact password are encrypted anew when they are being saved. The process to encrypt password is called `protect` in the jargon used by `libkeepass`. The problem is that the lib is not unprotecting the password fields in memory after they have been saved.

Unprotects the database means the passwords will be in plain text in memory. Unprotecting first is actually needed as the libkeepass `write_to()` method is encrypting the passwords before saving the file content. The `write_to()` method does not unprotect the fields after it has completed, meaning save operations that follow will encrypt again the already encrypted password fields, leading to a corrupted database (i.e. gibberish passwords).

### LINKED ISSUE #50: Use os.path.join
Use `os.path.join` instead of string manipulation in `tests.py` and elsewhere for crossplatform compatibility.

### LINKED ISSUE #83: Is installing the 'tests' module intentional?
When the package is installed, it installs the `tests` module because there is an `__init__.py` in the `tests/` directory and you use  `find_packages()`  in `setup.py`.

Just wondering if this is intentional? It seems like you shouldn't install tests with a production package, but I've seen it both ways with other packages.

Thanks!

Evidlo:
Changes
- integrated all the stuff from `libkeepass_proto` into `kdbx_parsing`
- eliminated `xmlfactory` and used the lxml element builder instead
- moved some common things from `entry.py` and `group.py` into `baseelement.py`
- pulled in some tests from `libkeepass`

Also, fixes for some issues:
- fixes #83
- fixes #50
- probably fixes #43 (I never had this issue, but all the old code has been replaced)
- issue #44 is irrelevant now


setup.py
```diff
@@ -5,15 +5,18 @@
...
     author='Philipp Schmitt',
     author_email='philipp@schmitt.co',
     url='https://github.com/pschmitt/pykeepass',
-    packages=find_packages(),
-    install_requires=['libkeepass', 'easypysmb', 'python-dateutil'],
-    entry_points={
-        'console_scripts': ['pkpwrite=pykeepass.pkpwrite:main']
-    }
+    packages=['pykeepass'],
+    install_requires=[
```
pschmitt: The libkeepass_proto dependency is missing here.

Evidlo: It's all in `kdbx_parsing/`

pschmitt: Oh okay - sorry, I should have actually read your description -_-'

+1 for deprecating libkeepass, in retrospect I should have done that a long time ago.

pschmitt: @Evidlo what features are missing ?

Evidlo: Twofish support and leaving entries protected (entries are always unprotected right now).

Also since the KDBX4 format shuffled things around, the location of certain low-level attribute changes.

For example, to access the protected stream key:

in KDBX3: `kp.kdbx.header.value.dynamic_header.protected_stream_key.data`
in KDBX4: `kp.kdbx.body.payload.inner_header.protected_stream_key.data`

Evidlo: Had any time to look at this?

Algram:
Can we merge this? It looks good to me and would make software relying on pykeepass usable for a lot of people again :+1: 

Thanks for the huge effort!

pschmitt:
Yes, we can ! :)

PR #87, opened 2018-07-01
Make PyKeePass usable as a context manager

JeffCarpenter:
A small patch so we can use `with` statements. Maybe it's no use to you if you're deprecating libkeepass /shrug

### DIFF
--- a/pykeepass/pykeepass.py
+++ b/pykeepass/pykeepass.py
@@ -27,6 +27,12 @@ class PyKeePass(object):
 
         self.read(password=password, keyfile=keyfile)
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, typ, value, tb):
+        self.kdb.close()
+
     def read(self, filename=None, password=None, keyfile=None):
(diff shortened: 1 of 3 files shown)


Evidlo:
Switching from `libkeepass` shouldn't have any affect on this.  The only change is that `kp.kdb` is renamed to `kp.kdbx`.

pykeepass/pykeepass.py
```diff
@@ -24,6 +24,12 @@ def __init__(self, filename, password=None, keyfile=None):
         self.kdb_filename = filename
         self.kdb = self.read(filename, password, keyfile)
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, typ, value, tb):
+        self.kdb.close()
```
pschmitt: Please change that, like @Evidlo mentioned.

JeffCarpenter:
Should be good to go now :+1:

pschmitt:
Great, thanks!

PR #93, opened 2018-08-27
Set UUID when creating new group

FalkAlexander:
When a group is created it needs also a UUID. Because the group creation procedure isn't located in the BaseElement module we need to do the UUID creation in the Group module itself. This prevents creating groups without an UUID.

### DIFF
--- a/pykeepass/group.py
+++ b/pykeepass/group.py
@@ -5,6 +5,8 @@ from lxml.etree import Element, _Element
 from lxml.objectify import ObjectifiedElement
 from lxml.builder import E
 import pykeepass.entry
+import base64
+import uuid
 
 
 class Group(BaseElement):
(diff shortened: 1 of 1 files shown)


Evidlo:
> Because the group creation procedure isn't located in the BaseElement module

What do you mean by this?  UUID creation is [here](https://github.com/pschmitt/pykeepass/blob/dac0c0bbeffdf8ff04cc5c47ddc642b81b4fd97d/pykeepass/baseelement.py#L22-L24).

FalkAlexander:
> What do you mean by this? UUID creation is here.

This piece of code doesn't get called for group creation but only for entry creation. Because of the if clause `if element is None:` bot element is not None when creating a new group.

Evidlo:
This is a mistake on my part.  I rewrote the `Group`, `Entry` and `BaseElement` init function a few weeks ago and I didn't think through the logic carefully enough.  I made a new PR that contains a proper fix: #94

pschmitt:
Fixed by #94

PR #96, opened 2018-08-29
customicon support

netson:
Updated to the latest master and reworked some parts to also support customicons for groups.

### DIFF
--- a/pykeepass/entry.py
+++ b/pykeepass/entry.py
@@ -2,7 +2,7 @@ from __future__ import unicode_literals
 from __future__ import absolute_import
 from pykeepass.baseelement import BaseElement
 from copy import deepcopy
-from lxml.etree import Element, _Element
+from lxml.etree import Element, _Element, iterwalk
 from lxml.objectify import ObjectifiedElement
 from lxml.builder import E
 import logging
@@ -17,6 +17,7 @@ reserved_keys = [
     'URL',
     'Tags',
     'IconID',
+    'CustomIconUUID',
     'Times',
     'History'
 ]
@@ -26,12 +27,13 @@ class Entry(BaseElement):
 
     def __init__(self, title=None, username=None, password=None, url=None,
                  notes=None, tags=None, expires=False, expiry_time=None,
-                 icon=None, element=None, version=None):
+                 icon=None, customicon=None, element=None, version=None, meta=None):
 
         assert type(version) is tuple, 'The provided version is not a tuple, but a {}'.format(
             type(version)
         )
         self._version = version
+        self._meta = meta
 
         if element is None:
             super(Entry, self).__init__(
@@ -39,7 +41,9 @@ class Entry(BaseElement):
                 version=version,
                 expires=expires,
                 expiry_time=expiry_time,
-                icon=icon
+                icon=icon,
+                customicon=customicon,
+                meta=meta
             )
             self._element.append(E.String(E.Key('Title'), E.Value(title)))
             self._element.append(E.String(E.Key('UserName'), E.Value(username)))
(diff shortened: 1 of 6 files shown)


pykeepass/entry.py
```diff
@@ -125,14 +129,6 @@ def notes(self):
     def notes(self, value):
         return self._set_string_field('Notes', value)
 
-    @property
-    def icon(self):
```
pschmitt: Why remove the `icon` getter and setters?

pykeepass/group.py
```diff
@@ -9,22 +9,24 @@
...
-                 version=None, expires=None, expiry_time=None):
+    def __init__(self, name=None, element=None, icon=None, customicon=None, notes=None,
+                 version=None, expires=None, expiry_time=None, meta=None):
 
         assert type(version) is tuple, 'The provided version is not a tuple, but a {}'.format(
             type(version)
         )
         self._version = version
-
+        self._meta = meta
```
pschmitt: Just curious: what's this `meta` stuff?

tests/tests.py
```diff
@@ -314,6 +314,7 @@ def test_set_and_get_fields(self):
         entry.expires = False
         entry.expiry_time = changed_time
         entry.icon = icons.GLOBE
+        entry.customicon = "9"
```
pschmitt: Should this even be valid? I was expecting more something like an actual image file.

tests/tests.py
```diff
@@ -391,12 +393,6 @@ def test_db_info(self):
     def tearDown(self):
         os.remove(os.path.join(base_dir, 'change_creds.kdbx'))
 
-class CtxManagerTests(unittest.TestCase):
```
pschmitt: Why remove this?

pykeepass/entry.py
netson: because the exact same functions exist in the baseelement; didn't seem necessary

pykeepass/group.py
netson: The KDBX XML contains a Meta tag which contains the customicons (the actual image files). Since it doesn't use a generic key index but instead a base64 index to link the image to the entry, I send the meta tag + children to the the entry, so the base64 index can be converted to a numerical one (as the keepass UI does as well)

tests/tests.py
netson: The way keepass works is that when you set a customicon, it sets the default icon to 0. If you do not want to use the customicon anymore, the CustomIconUUID tag should be removed from the entry. I can only create a separate test for customicons if I modify the test KDBX files; the code will not set customicon to a value which does not actually exist. So setting this to 9 will only result in a customicon with index 9 if at least 10 customicons exist. In all other cases (as it does here) customicon is set to None, which removes the tag from the entry XML.

tests/tests.py
netson: I don't think this was intentional, but probably happened because you merged some pull requests after I rebased; I guess it can be ignored.

tests/tests.py
netson: to elaborate, customicons are only stored in the database once and are referred to by entries by using a UUID which is base64 encoded. This PR does not add the possibility to add actual image files as custom icons, but merely allows you to set customicons which already exist in the DB by providing their numerical index.

Elias481:
I would also really like custom icons.
But for me the icon-management part (add, remove, lookup) similiar to binary attachment would be important, too.
And I dislike the idea to work with indices here as opposed to the attachment the custom icon do not have indices but UUIDs. So why intoduce an index here that does not really exists (and by definition is not stable as XML allows to reorder the elements)?
(I would have no issues with making the SETTER flexible, so allowing for index and the actual binary data optionally but converting this to the UUID then - and have an icons-function that resolves the UUID to binary - and possibly the current index.)

I see it's long time this PR is pending now...

tests/tests.py
shining-fnml: Take a look at the PR I've just pushed. https://github.com/libkeepass/pykeepass/pull/417

shining-fnml:
@Elias481 
> I would also really like custom icons. [...] And I dislike the idea to work with indices here as opposed to the attachment the custom icon do not have indices but UUIDs. [...]

Take a look at the PR I've just pushed: https://github.com/libkeepass/pykeepass/pull/417

PR #102, opened 2018-10-08
Decode Database After Calling Save

FalkAlexander:
When calling the save function the protected elements of the database tree become encoded again. That means that all passwords are encrypted and not user-readable. This is bad if we want to continue to work on the database after saving instead of instantly closing it. That means all passwords are gibberish after the save. In order to prevent this we reparse the database to decode the protected elements again.

### DIFF
--- a/pykeepass/pykeepass.py
+++ b/pykeepass/pykeepass.py
@@ -59,6 +59,12 @@ class PyKeePass(object):
                 )
             )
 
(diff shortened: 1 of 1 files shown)


pschmitt:
Thanks for the contribution. Could you provide a snippet so that I can verify that this works?

Evidlo:
Fixed in 3.2.0

Issue #112, opened 2018-11-04
save_history only working once


BoomerB:
Hi,

I tried to add history to my entries

I'm using Python 3.6.3 on RHEL7.4
I'm using pykeepass (3.0.2) 

**1st run**
```
from pykeepass import PyKeePass
kp = PyKeePass('testing.kdbx', 'secret')
group = kp.find_groups(name='python',  first=True)
item = kp.find_entries(group=group, title='hostname',username='root',first=True)
item.save_history()
kp.save()

```

so far so good

**2nd run**
```
from pykeepass import PyKeePass
kp = PyKeePass('testing.kdbx', 'secret')
group = kp.find_groups(name='python',  first=True)
item = kp.find_entries(group=group, title='hostname',username='root',first=True)
print(item.history)
[b'Entry: "[History of: hostname] (root)"']
item.save_history()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/pykeepass/entry.py", line 229, in save_history
    archive.remove(archive.History)
AttributeError: 'lxml.etree._Element' object has no attribute 'History'

```

BoomerB:
and `touch `seems to have a similar problem:

```
from pykeepass import PyKeePass
kp = PyKeePass('/var/www/webdav/Linux_DUB.kdbx', '10AUmdBk23')
group = kp.find_groups(name='Robot',  first=True)
item = kp.find_entries(group=group, title='gugus',username='root',first=True)
>>> print(item.atime)
2018-11-04 13:52:03.902889+00:00
>>> print(item.mtime)
2018-11-04 13:52:03.902889+00:00
>>> print(item.ctime)
2018-11-04 13:52:03.902889+00:00
item.touch()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/pykeepass/entry.py", line 219, in touch
    self._element.Times.LastAccessTime = datetime.utcnow()
AttributeError: 'lxml.etree._Element' object has no attribute 'Times'

```

pschmitt:
@BoomerB Thanks. Can you please a PR with said change?

BoomerB:
Pull Request #113

PR #113, opened 2018-11-06
fix "object has no attribute" issue in save_history

BoomerB:
Not sure if this is the right way to fix this issue but it solved my problem with python2 and python3. maybe there's a way to tell lxml to create the attributes when loading the file.


BoomerB:
see issue #112

pschmitt:
Thanks !

pschmitt:
Thanks !

PR #114, opened 2018-11-09
fix "object has no attribute" for touch

BoomerB:
the class BaseElement has methods to set Times, so we should use them


whwright:
Why not add some tests? They are listed as missing tests here https://github.com/pschmitt/pykeepass/blob/f10766289f0ab6cb3f155488fb9b6ba03048acab/tests/tests.py#L19

pschmitt:
@BoomerB Can you please rebase and include #121 ?

Evidlo:
@BoomerB bump

If you're not comfortable rebasing, you could close this and #121 and make a new PR.

PR #116, opened 2018-11-11
Fix datetime comparison when there is no tzinfo and autotype sequence none check

whwright:
Resolves issues: https://github.com/pschmitt/pykeepass/issues/115 and https://github.com/pschmitt/pykeepass/issues/117

### DIFF
--- a/pykeepass/baseelement.py
+++ b/pykeepass/baseelement.py
@@ -78,7 +78,7 @@ class BaseElement(object):
 
     def _datetime_to_utc(self, dt):
         """Convert naive datetimes to UTC"""
-        
+
         if not dt.tzinfo:
             dt = dt.replace(tzinfo=tz.gettz())
         return dt.astimezone(tz.gettz('UTC'))
@@ -150,7 +150,9 @@ class BaseElement(object):
 
     @property
     def expired(self):
-        return self.expires and (datetime.utcnow() > self.expiry_time)
+        if self.expires:
+            return self._datetime_to_utc(datetime.utcnow()) > self._datetime_to_utc(self.expiry_time)
+        return False
 
 
     @property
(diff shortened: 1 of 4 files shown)


Evidlo:
I think pschmitt likes to do the version bumping himself.

Could we get tests added for these two issues?

Also, you might be interested in [passhole](https://github.com/Evidlo/passhole), since you seem to be working on a Keepass CLI tool.

whwright:
Test! FWIW I did some TDD by reverted my changes, wrote tests, made sure they blew up, and added my changes back.

Also `passhole` looks awesome, but the focus of my tool is very narrow for something I do in a personal workflow.

Evidlo:
Could you use `kp.add_entry` instead of instantiating those entries yourself?

Also, you can use `git commit --amend --no-edit` and `git push -f` to make changes without adding new commits to your PR.

Evidlo:
That's because I wrote `EntryTests` to be independent from `add_entry` failures.  All other tests that create entries use `add_entry`.  Using `Entry` directly is discouraged because it tends to change often.

whwright:
I think it makes sense to instantiate the `Entry` object directly to isolate this unit test as much as possible. `add_entry` should be tested separately, which it is. You will have to update the `EntryTests` if the fundamentals of the objects change anyway.

The `add_entry` function also doesn't pass through an `element` property, which I am using in my test. Again it makes sense to me to do the testing at this level, and avoid any further re-factoring or munging with the private instance properties.

What is your opinion @pschmitt?

whwright:
I have opened up a discrete issue for the auto type error, instead of having it piggyback here.

Evidlo:
Looks good to me.

whwright:
Awesome thank you! When do you typically do releases?

PR #134, opened 2019-01-21
Update setup.py

firecat53:
Minor bugfix. Package won't build otherwise


Evidlo:
Damn, that's my fault.  We should make Travis test this.

PR #135, opened 2019-01-22
Autotype property fix

firecat53:
The property `autotype_sequence` was not returning a value when a value was clearly visible using dump_xml() or viewing the autotype sequence in KeepassXC. This PR should fix that problem. All tests still pass. Included one small related PEP8 fix as well. Thanks!

### DIFF
--- a/pykeepass/entry.py
+++ b/pykeepass/entry.py
@@ -195,7 +195,7 @@ class Entry(BaseElement):
     def autotype_enabled(self, value):
         enabled = self._element.find('AutoType/Enabled')
         if value is not None:
-            enabled.text= str(value)
+            enabled.text = str(value)
         else:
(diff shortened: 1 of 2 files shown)


Evidlo:
Could you modify tests.py so that it catches the problem?

firecat53:
Sorry, forgot to include the warning received (Python 3.7):

    /home/firecat53/~/projects/pykeepass/pykeepass/entry.py:206: FutureWarning: The behavior of this 
    method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.
    return sequence.text if sequence else None

firecat53:
I'll have a look at tests.py

firecat53:
That should do it, I think. Thanks for the prompt!

PR #136, opened 2019-01-27
Make 'path' ignore other arguments.

### LINKED ISSUE #127: Behavior of 'path'
What should `path` do?

Initially we intended for it to be a way to directly select a *single* entry or group, but later on its behavior was modified to also mean the path to search under (this was befor `group` was added).

It doesn't make sense to do both and it makes `_find()` a pain, so we should pick one behavior and stick with it..

I'm of the opinion that when `path` is present, all other search terms should be ignored and the entry/group returned directly (like the original behavior of `find_x_by_path`).  People can use `group` to do searches under a specific element.

``` python
>>> kp.find_entries(path='foobar_group/foobar_entry', title='blah', password='blah')
Entry: "foobar_group/foobar_entry (foobar_user)"
```

And I suppose `path` implies `first=True`.

Evidlo:
This is what I proposed in #127.  It also fixes a bug where `find_entries('foobar_group/sub_group/', first=True)` could return a `Group` instead of an `Entry`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -275,12 +275,12 @@ the data that this attachment points to.  Raises ``AttachmentError`` if data doe
    >>> e = kp.add_entry(kp.root_group, title='foo', username='', password='')
 
    # add attachment data to the db
-   >>> attachment_id = kp.add_binary(b'Hello world')
-   >>> kp.attachments
+   >>> binary_id = kp.add_binary(b'Hello world')
+   >>> kp.binaries
    [b'Hello world']
 
    # add attachment reference to entry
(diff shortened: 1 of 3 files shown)


Evidlo:
Hold off on merging this.  I need to correct something in `_find`.

Evidlo:
Should be good now.

Evidlo:
Any news on this?

Evidlo:
Just wondering if there is anything keeping this from being merged?

Evidlo:
Bump

PR #145, opened 2019-03-08
zero-index attachment id

Evidlo:
- `kp.add_attachment` is now zero indexed so that it can be used directly in `kp.binaries`
- add missing function documentation for `Attachment.entry`
- rename `AttachmentError` to `BinaryError` and fix broken import. add test
- alias `Attachment.binary` to `Attachment.data`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -270,6 +270,10 @@ string representing this attachment
 
 the data that this attachment points to.  Raises ``AttachmentError`` if data does not exist.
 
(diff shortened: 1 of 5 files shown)


Evidlo:
Heads for for @cristim and @zioalex, that this is a breaking change.

Evidlo:
Moved to https://github.com/pschmitt/pykeepass/pull/147

Issue #143, opened 2019-03-01
attachment.data returns the oldest content, not the latest


sergiofagostinho:
Hi,

We've found the following code:
```python
...
for entry in entries:
  ...
     attachment = entry.attachments[0]
     secret = attachment.data
```

only works when you delete the history.

Doesn't matter if you delete the attachment or add it with a different name.

I've attached a test file (password: `password`) where I've added a PNG file, deleted it and added a text file. If you run the code it will still display the binary data.

kbdx file: https://drive.google.com/open?id=1ms3Jobn_rxfDTrmK2bE2x6Mn3OZzxIMK

Also, I'm writing this file with the latest version of KeePassX on Mac (2.0.3).

Evidlo:
Could you provide a more complete example of what you mean?  Everything seems to work as expected for me.

sergiofagostinho:
Hi @Evidlo. I would expect the `attachment.data` to hold the current file content (text), not the file I've deleted (PNG image). Basically, what I see in Keepass X:

![image](https://user-images.githubusercontent.com/2901192/54019004-0069d300-4182-11e9-880c-ff70cde1fe81.png)

pscosta:
I'm able to reproduce the issue calling the following method:

`find_attachments (id=None, filename=None, element=None, recursive=True, regex=False, flags=None, history=False, first=False)`

even thought I set `history=False` I always get old  attachments from the history.
The workaround for me is to delete all history entries before retrieving attachments.

Evidlo:
Fixed by PR #147

PR #147, opened 2019-03-17
Fix <= version (4,0) ~version 3 add_binary ID referencing

disarticulate:
I inspected a kdbx created by KeePass, and found that /KeePassFile/Meta/Binaries/Binary requires an ID key to reference the proper data, and fixed a `data = base64.b64encode(data).decode` typo which was not decoding the data.

See https://github.com/pschmitt/pykeepass/issues/146#issue-421943414 for debug details

### DIFF
--- a/pykeepass/entry.py
+++ b/pykeepass/entry.py
@@ -26,7 +26,8 @@ reserved_keys = [
     'Tags',
     'IconID',
     'Times',
-    'History'
+    'History',
+    'Notes'
 ]
 
 # FIXME python2
--- a/pykeepass/entry.py
+++ b/pykeepass/entry.py
@@ -195,14 +195,14 @@ class Entry(BaseElement):
     def autotype_enabled(self, value):
         enabled = self._element.find('AutoType/Enabled')
         if value is not None:
-            enabled.text= str(value)
+            enabled.text = str(value)
         else:
             enabled.text = None
 
     @property
     def autotype_sequence(self):
         sequence = self._element.find('AutoType/DefaultSequence')
-        return sequence.text if sequence else None
+        return sequence.text if sequence is not None else None
 
     @autotype_sequence.setter
     def autotype_sequence(self, value):
--- a/pykeepass/pykeepass.py
+++ b/pykeepass/pykeepass.py
@@ -494,14 +494,14 @@ class PyKeePass(object):
                 # gzip compression
                 data = zlib.compress(data)
             data = base64.b64encode(data).decode()
+
             # set ID for Binary Element
             ID = str(len(self.binaries) + 1)
-            
+
             # add binary element to XML
             binaries.append(
                 E.Binary(data, ID=ID, Compressed=str(compressed))
             )
-            #print(len(self.binaries), self._xpath('/KeePassFile/Meta/Binaries'), len(binaries))
 
         # return attachment id
         return len(self.binaries)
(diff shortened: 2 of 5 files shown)


Evidlo:
Please wait until https://github.com/disarticulate/pykeepass/pull/1 is merged before merging this.

lmussier:
Hi,

Any update  / ETA on this PR.
We are facing the exact same issue  :s

sergiofagostinho:
Also looking forward for this, as we suspect this will address #143.

Venti-:
I tried this patch, specifically https://github.com/disarticulate/pykeepass/commit/e6f73e7623f5cd7b3f157a70f06e4f27f5024dec, and found that this fixes some of the issue, but weirdness remains.

- After adding an attachment, PyKeePass.attachments is None, and the attachment is only seen in binaries.
- With a kdbx3.1 file, I could not open the attachment with Strongbox or MacPass. In StrongBox I could see the attachment, but the content was not decrypted correctly. In MacPass I could not see the attachment at all. PyKeePass could read the attachment.
- With a kdbx4.0 file, saving the attachment seemed to work correctly.

I don't know enough about KeePass file format to figure out what's wrong exactly, but here is the code to reproduce the problem.

```python
kp = PyKeePass(KP_PATH, password=password)

# Setup one group named test_group without any entries.
while len(kp.binaries) > 0:
    kp.delete_binary(0)
group = kp.find_groups(path='test_group', first=True)
if group is None:
    group = kp.add_group(kp.root_group, 'test_group')
entry = kp.find_entries(group=group, title='test_title', first=True)
if entry is not None:
    kp.delete_entry(entry)

print('# attachments', kp.attachments)
print('# binaries', kp.binaries)

entry = kp.add_entry(kp.root_group, title='test_title', username='', password='')
attachment_id = kp.add_binary(b'test bytes')
entry.add_attachment(attachment_id, 'test_filename')

print("# entry.attachments =", entry.attachments)
attachment = entry.attachments[0]
print("# attachment.id='{id}', attachment.filename='{filename}', attachment.data='{data}'".format(
    id=attachment.id, filename=attachment.filename, data=attachment.data))

print('# attachments', kp.attachments)
print('# binaries', kp.binaries)

kp.save()
```
```
# version (3, 1)
# kdf_algorithm aeskdf
# binaries [b'test bytes']
# kp.delete_binary(0)
# attachments None
# binaries []
# entry.attachments = [Attachment: 'test_filename' -> 0]
# attachment.id='0', attachment.filename='test_filename', attachment.data='b'test bytes''
# attachments None
# binaries [b'test bytes']
```

Evidlo:
Don't merge this until https://github.com/disarticulate/pykeepass/pull/2 is merged

Evidlo:
@disarticulate Could you merge https://github.com/disarticulate/pykeepass/pull/2

disarticulate:
oops. sorry guys, i thought we merged em together.

Garcia6l20:
@disarticulate Maybe you could re-open ?

Garcia6l20:
@disarticulate I might have warned you, I already had same issue. Try [this](https://github.com/pschmitt/pykeepass/pull/166/files).

Evidlo:
@disarticulate Why again are we doing `.decode()` after `b64decode`?  Shouldn't the result be a bytestring, not a regular string, since these are binaries?

disarticulate:
All I knew was decode is a function which doesn't make sense in the context.

It could have been it's own merge request.

PR #165, opened 2019-08-30
Fix/attachments

Garcia6l20:
Fixes some keepass2 GUI attachments interoperabilities:
- missing ID attribute in *KeePassFile/Meta/Binaries/Binary* tags
- `kp.add_binary` returned wrong id
- using gzip to (de)compress binaries


Garcia6l20:
Closing PR for python2 compat work

Issue #120, opened 2018-12-01
entry.touch() doesn't work


BoomerB:
```
$ python
Python 2.7.15rc1 (default, Nov 12 2018, 14:31:15) 
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from pykeepass import PyKeePass
>>> kp = PyKeePass('test.kdbx', 'monilove')
>>> group_name = 'Robot'
>>> entry_name = 'KPserver'
>>> group = kp.find_groups(name=group_name, first=True)
>>> item = kp.find_entries(group=group, title=entry_name, first=True)
>>> item.touch(modify=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/pykeepass/entry.py", line 219, in touch
    self._element.Times.LastAccessTime = datetime.utcnow()
AttributeError: 'lxml.etree._Element' object has no attribute 'Times'
```

BoomerB:
I submitted a pull request some time ago
#114

Evidlo:
Please add a test.

Evidlo:
Did you ever make a PR for this?  It seems like they're all closed.

BoomerB:
Hi,
I tried several times but it never really worked. last time I tried to
contribute, I wanted to run the checks on my changes before I create a pull
request but all tests failed for some other changes. I waited for a while
in the hopes that the one who caused the problems would fix them and then I
was busy with other things.
I think I still have the code somewhere. I can try to organize it into a
few small pull requests. Then it might be easier to get them through

Kind regards
Bernhard

Am Do., 12. Sept. 2019 um 04:05 Uhr schrieb Evan Widloski <
notifications@github.com>:

PR #170, opened 2019-09-15
changed entry.touch() to use member functions to change atime/mtime

BoomerB:
https://github.com/libkeepass/pykeepass/issues/120 is open for some time and I almost forgot about it
here's another try to get this fixed

I made 2 changes:
- I changed the code for the touch function to use the atime.setter/mtime.setter from the BaseElement class
- I added a test in the EntryTests class to test the touch function. Sorry for the sleep statement but I had no better idea

Hope this helps
BoomerB

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -16,7 +16,6 @@ import os
 import shutil
 import unittest
 import logging
-import time
 
 """
 Missing Tests:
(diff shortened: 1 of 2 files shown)


tests/tests.py
```diff
@@ -375,6 +376,33 @@ def test_expired_datetime_offset(self):
...
+        entry = Entry(
+            'title',
+            'username',
+            'password',
+            kp=self.kp
+        )
+        atime = entry.atime
+        mtime = entry.mtime
+        ctime = entry.ctime
+        time.sleep(1)
```
Evidlo: I think `atime`, `ctime`, and `mtime` all record microseconds, so can't we get away with a much smaller sleep?

BoomerB:
I tested with 0.5s and it's not stable, most of the time it passes but sometimes EntryTests4 fails on the atime assertion.
I could change the test and use an existing entry for touch. then we don't need to wait because the entry is already much older.

Evidlo:
Yes I think that would be better.

BoomerB:
BTW: I noticed the following:
```
>>> from pykeepass import PyKeePass
>>> database = 'test3.kdbx'
>>> password = 'password'
>>> keyfile = 'test3.key'
>>> kp = PyKeePass(database, password=password, keyfile=keyfile)
>>> 
>>> from datetime import datetime, timedelta
>>> from dateutil import tz
>>> from pykeepass.entry import Entry
>>> now = datetime.now(tz.gettz())
>>> utc = datetime.now(tz.gettz('UTC'))
>>> entry = Entry(
...     'title',
...     'username',
...     'password',
...     kp=kp
... )
>>> 
>>> now
datetime.datetime(2019, 9, 19, 23, 0, 7, 382284, tzinfo=tzfile('/etc/localtime'))
>>> utc
datetime.datetime(2019, 9, 19, 21, 0, 7, 382512, tzinfo=tzfile('/usr/share/zoneinfo/UTC'))
>>> entry.atime
datetime.datetime(2019, 9, 19, 19, 0, 7, 383109, tzinfo=tzfile('/usr/share/zoneinfo/UTC'))
```
there are 2 hours between localtime and UTC and another 2 hours between utc and atime
so, it looks like time is converted twice

BoomerB:
changed test to use existing entries to get rid of the sleep statements

BoomerB:
BTW: if you have a look at https://github.com/libkeepass/pykeepass/pull/176 you can see that only KDBX V3 stores microseconds but V4 stores only seconds. that's why only sleep(1) will work and sleep(0.X) will be unstable.

Issue #174, opened 2019-09-16
entry.history


BoomerB:
Hi,
I'm working on some tests for the history feature
now I noticed different behavior for entries in the test files and entries created by pykeepass
the entries in the test files always return a list, e.g. when the entry has no history it returns an empty list
entries created by pykeepass return None when there's no history because there's no "History" tag in the XML data
I assume the test files were created by the original KeePass for maximum compatibility.
So, now the question is: is this deviation acceptable? or should we change the behavior of pykeepass?
Kind regards
Boomer

Evidlo:
Are you referring to `Entry.history`?  If so, I think it should return an empty list when there are no history items, regardless of the XML.

BoomerB:
ok, this is easy

Issue #137, opened 2019-01-29
regression in master: AttributeError: 'PyKeePass' object has no attribute 'kdbx'


singuliere:
The following

    entries = []
    with PyKeePass(filename, password=password, keyfile=keyfile) as kp:
         entries.extend(kp.entries)
    entries[0].expiry_time

raises an exception as of 193b39177f79d2d616daab5c7da3a3146f725986:

    File "pykeepass/pykeepass/baseelement.py", line 179, in expiry_time
     return self._get_times_property('ExpiryTime')
    File "pykeepass/pykeepass/baseelement.py", line 149, in _get_times_property
      return self._decode_time(prop.text)
    File "pykeepass/pykeepass/baseelement.py", line 123, in _decode_time
      if self._kp.version >= (4, 0):
    File "pykeepass/pykeepass/pykeepass.py", line 83, in version
      self.kdbx.header.value.major_version,

because the kdbx data member is [removed by \_\_exit__](https://github.com/pschmitt/pykeepass/blob/193b39177f79d2d616daab5c7da3a3146f725986/pykeepass/pykeepass.py#L49). This regression was introduced in [the attachment implementation](https://github.com/pschmitt/pykeepass/pull/126) which [keeps a reference to the PyKeePass object](https://github.com/pschmitt/pykeepass/pull/126/files#diff-c95e006146be0e4b924d2ee431dd1b8dR89) and relies on its implementation of the **version** attribute which requires the **kdbx** data member.

singuliere:
@Evidlo here is one use case: https://github.com/pschmitt/vault-keepass-import/blob/master/keepass_import.py#L89 (using libkeepass) which is converted to using pykeepass at https://github.com/singuliere/vault-keepass-import/blob/update/vault_keepass_import/main.py#L35

However, even if there was no known use case it would still be a regression that needs fixing, don't you think?

Evidlo:
> However, even if there was no known use case it would still be a regression that needs fixing, don't you think?

We've always assumed the database will be open while accessing entries/groups, so I think this falls under 'undefined behavior'.  You should convert the entries into primitives while inside the context manager.

singuliere:
Would you be open to the idea of a pull request implementing *copy* method for the entry that would return a copy of the object that does not rely on the underlying database? I could implement such a function in my own code. But it looks like something other people could use and better implemented in pykeepass itself. I may not be the only one stumbling on this unexpected behavior.

singuliere:
[...]

PR #175, opened 2019-09-17
History tests

BoomerB:
Hi,
here some tests for the history feature
as discussed in https://github.com/libkeepass/pykeepass/issues/174 I also changed the code Entry.history to make sure that it always returns a list
please review and let me know what you think. I know the tests are not perfect but mostly serve as a proof-of-concept
Kind regards
Boomer

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -136,6 +136,48 @@ class EntryFindTests3(KDBX3Tests):
         results = self.kp.find_entries(title='foobar_entry', group=group)
         self.assertEqual(len(results), 2)
 
+    #---------- History -----------
+
+    def test_is_a_history_entry(self):
+        for title in ["root_entry", "subentry"]:
+            res1 = self.kp.find_entries_by_title(title)
+            for entry in res1:
+                self.assertFalse(entry.is_a_history_entry)
+            res2 = self.kp.find_entries_by_title(title, history=True)
+            self.assertTrue(len(res2) > len(res1))
+            for entry in res2:
+                if entry not in res1:
+                    self.assertTrue(entry.is_a_history_entry)
+
+    def test_history(self):
+        entry = self.kp.find_entries_by_title("subentry2", first=True)
+        hist = entry.history
+        self.assertIsInstance(hist, list)
+        self.assertEqual(len(hist), 0)
+
+        entry = self.kp.find_entries_by_title("subentry", first=True)
+        hist = entry.history
+        self.assertIsInstance(hist, list)
+        self.assertEqual(len(hist), 4)
+
+    def test_history_path(self):
+        for title in ["root_entry", "subentry"]:
+            entry = self.kp.find_entries_by_title(title, first=True)
+            hist = entry.history
+            self.assertTrue(len(hist) > 0)
+            for item in hist:
+                self.assertEqual(item.path, '[History of: {}]'.format(entry.title))
+
+    def test_history_group(self):
+        for title in ["root_entry", "subentry"]:
+            entry = self.kp.find_entries_by_title(title, first=True)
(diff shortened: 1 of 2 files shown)


tests/tests.py
```diff
@@ -407,6 +449,158 @@ def test_add_remove_attachment(self):
...
+                backup[entry.uuid] = {"atime": entry.atime, "mtime": entry.mtime, "ctime": entry.ctime}
+                entry.title = changed + 'title'
+                entry.username = changed + 'user'
+                entry.password = changed + 'pass'
+                entry.atime = now
+                entry.mtime = now
+            kp2.save()
+
+        # changed entries tests
+        with PyKeePass(fn, password=self.password, keyfile=keyfile) as kp2:
```
Evidlo: I would prefer to avoid using the context manager in tests.  It doesn't actually do anything any more so we should discourage its use: https://github.com/libkeepass/pykeepass/issues/137

tests/tests.py
```diff
@@ -407,6 +449,158 @@ def test_add_remove_attachment(self):
...
+    def test_find_history_entries(self):
+        '''run some tests on entries created by pykeepass'''
+        prefix = 'TFE_'
+        postfix = '_tmpA'
+        changed = 'tfe_changed_'
+        # create a new copy of the file to ensure the original file remains unchanged
+        a = self.kp.filename.split('.')
+        fn = '.'.join(('.'.join(a[:-1]) + postfix, a[-1]))
+        keyfile = os.path.join(base_dir, self.keyfile)
+        self.kp.save(filename=fn)
```
Evidlo: Why is it necessary to make a new database for these tests?  Can't you use the existing test database and just not save it?

tests/tests.py
```diff
@@ -407,6 +449,158 @@ def test_add_remove_attachment(self):
...
+            self.assertEqual(len(res1), 3)
+            for entry in res1:
+                self.assertFalse(entry.is_a_history_entry)
+                hist = entry.history
+                self.assertEqual(len(hist), 1)
+                for item in hist:
+                    self.assertTrue(item.is_a_history_entry)
+                    self.assertEqual(item.group, entry.group)
+                    self.assertEqual(item.path, '[History of: {}]'.format(entry.title))
+                    # check if everthing was copied as expected
```
Evidlo: Under the hood, `save_history()` just duplicates the Entry and appends it to history, so it's probably not necessary to test every entry attribute here.  We'd just be writing tests for lxml's `.append()` function.

tests/tests.py
BoomerB: I like context managers because they cleanup after me and I don't have to think about closing files and stuff

tests/tests.py
BoomerB: I noticed the for entries in the test database the "History" element is always there but sometimes empty but for entries created by pykeepass the "History" element is missing for entries with no history. My first assumption was that the kp.save() function applies magic to add the "History" element when missing. Therefore I decided to save the DB and reload it.
Now that we know that there's no magic, we can skip this test

BoomerB:
updated the request, hope I followed all yours

tests/tests.py
Evidlo: The file is closed immediately after PyKeePass is finished reading the database.

Evidlo:
Seems to be missing `EntryHistoryTests4`.

Evidlo:
On closer inspection, I think some of these tests need to be reorganized.  For example, `test_history_path`, `test_history_group` and `test_is_a_history_entry` really belong in  `EntryTests3`.

Also, I think I'd like to merge `EntryTests3` and `EntryHistoryTests3`.  I'll submit a PR.

Issue #108, opened 2018-10-30
Feature request: Support field references


philipsd6:
```python
In [9]: kp.find_entries(title='Test Entry', first=True).password
Out[9]: '{REF:P@I:49EF54B695EB3E4591F3F07D26D4124C}'
```
Ideally, this would automatically dereference that and return the password, or there should be methods to track it down, like maybe:
```python
kp.deref(entry.password)
```
Ultimately, there should also be a method to link one entry's field to another entry's field, perhaps like:
```python
existing_entry = kp.find_entries(title='Production Entry', first=True)
url = 'https://test-site.example.com/?user=%s' % existing_entry.ref('username')
new_entry = kp.add_entry('Testing',
                         title='Test Entry',
                         username=existing_entry.ref('username'),
                         password=existing_entry.ref('password'),
                         url=url)
```

philipsd6:
On a related note, why are the uuids not in the same format as in Keepass2? Trying to do this manually doesn't work:
```python
In [33]: kp.find_entries(uuid='49EF54B695EB3E4591F3F07D26D4124C')
Out[33]: []
```
But this works:
```python
In [34]: kp.find_entries(uuid='Sd9UpoArPkWA4/C/F9QZSw==')
Out[34]: [b'Entry: "Work/Production Entry (testuser)"']
```
How to convert from one uuid format to the other, and why should the end user do that?

Evidlo:
The way you convert is with `base64.b64encode()` and `base64.b64decode()`.  They are base64 encoded because that's how they are stored in XML.

I'll look into converting it automatically in the next few days.

Evidlo:
I don't understand your first post.  Is `{REF:P@I:49EF54B695EB3E4591F3F07D26D4124C}` some string being generated by your password manager? That's definitely not anything on our side.

philipsd6:
@Evidlo That string is standard Keepass2 field reference stuff: https://keepass.info/help/base/fieldrefs.html

In the Windows Keepass2 app, `{REF:P@I:49EF54B695EB3E4591F3F07D26D4124C}` returns the **P**assword from the entry referenced by the given UU**I**D

Evidlo:
Support added by #178

PR #178, opened 2019-10-14
Support for field references.

mats-erik:
This adds support for KeePass standard field references https://keepass.info/help/base/fieldrefs.html
- deref() for replacing a reference with value of referenced attribute of another entry or group. Also handles recursion as the referenced attribute may itself refer to another.
- ref() for creating a reference.


mats-erik:
This would close issue: https://github.com/libkeepass/pykeepass/issues/108

Evidlo:
Thanks, but it needs tests before it can be merged.

PR #186, opened 2019-12-17
Allow setting protected custom properties

jeandeaual:
- I added a new argument to `Entry.set_custom_property` to set protected custom properties.

    It can be used as follows:

    ```python
    entry.set_custom_property('Key', 'Protected Value', protect=True)
    ```

- I also noticed that when creating a new entry, the `protected` attribute on the password value tag was lowercase, while it appears that it should start with an uppercase (see https://github.com/keepassxreboot/keepassxc-specs/blob/master/kdbx-xml/rfc.md#protected-elements).

    I verified that by creating a new entry using KeePass and looking at the generated XML. We can see that the password is generated as `<Key>Password</Key><Value Protected="False">NewPass</Value>`:

    ```
    >>> from pykeepass import PyKeePass
    >>> kd = PyKeePass('test.kdbx', password='pass')
    >>> e = kd.find_entries(title='Test', first=True)
    >>> from lxml import etree
    >>> etree.tostring(e._element)
    b'<Entry>\n\t\t\t\t\t<UUID>eYM06IgGl0eVDBYo+Fsnkw==</UUID>\n\t\t\t\t\t<IconID>18</IconID>\n\t\t\t\t\t<ForegroundColor/>\n\t\t\t\t\t<BackgroundColor/>\n\t\t\t\t\t<OverrideURL/>\n\t\t\t\t\t<Tags/>\n\t\t\t\t\t<Times>\n\t\t\t\t\t\t<CreationTime>Bc2K1Q4AAAA=</CreationTime>\n\t\t\t\t\t\t<LastModificationTime>Ps2K1Q4AAAA=</LastModificationTime>\n\t\t\t\t\t\t<LastAccessTime>Ps2K1Q4AAAA=</LastAccessTime>\n\t\t\t\t\t\t<ExpiryTime>ycyK1Q4AAAA=</ExpiryTime>\n\t\t\t\t\t\t<Expires>False</Expires>\n\t\t\t\t\t\t<UsageCount>1</UsageCount>\n\t\t\t\t\t\t<LocationChanged>Bc2K1Q4AAAA=</LocationChanged>\n\t\t\t\t\t</Times>\n\t\t\t\t\t<String>\n\t\t\t\t\t\t<Key>Notes</Key>\n\t\t\t\t\t\t<Value/>\n\t\t\t\t\t</String>\n\t\t\t\t\t<String>\n\t\t\t\t\t\t<Key>Password</Key>\n\t\t\t\t\t\t<Value Protected="False">NewPass</Value>\n\t\t\t\t\t</String>\n\t\t\t\t\t<String>\n\t\t\t\t\t\t<Key>Protected field</Key>\n\t\t\t\t\t\t<Value Protected="False">Protected value</Value>\n\t\t\t\t\t</String>\n\t\t\t\t\t<String>\n\t\t\t\t\t\t<Key>Standard field</Key>\n\t\t\t\t\t\t<Value>Standard value</Value>\n\t\t\t\t\t</String>\n\t\t\t\t\t<String>\n\t\t\t\t\t\t<Key>Title</Key>\n\t\t\t\t\t\t<Value>Test</Value>\n\t\t\t\t\t</String>\n\t\t\t\t\t<String>\n\t\t\t\t\t\t<Key>URL</Key>\n\t\t\t\t\t\t<Value/>\n\t\t\t\t\t</String>\n\t\t\t\t\t<String>\n\t\t\t\t\t\t<Key>UserName</Key>\n\t\t\t\t\t\t<Value>User</Value>\n\t\t\t\t\t</String>\n\t\t\t\t\t<AutoType>\n\t\t\t\t\t\t<Enabled>True</Enabled>\n\t\t\t\t\t\t<DataTransferObfuscation>0</DataTransferObfuscation>\n\t\t\t\t\t</AutoType>\n\t\t\t\t\t<History/>\n\t\t\t\t</Entry>\n\t\t\t'
    ```


Evidlo:
I don't quite understand why you want this feature.  In KeePass and KeePassXC, they use this attribute to keep track of which entry passwords are encrypted in memory (and the decrypt key is kept in secure memory somewhere).  Since Python's memory [can't be secure](https://stackoverflow.com/questions/728164/securely-erasing-password-in-memory-python), we don't really use this.  

All entry passwords are unprotected as one of the steps in decrypting the database: https://github.com/libkeepass/pykeepass/blob/master/pykeepass/kdbx_parsing/common.py#L174-L199

jeandeaual:
I understand that pykeepass can't make use of this attribute when opening a database.
I just want to be able to use pykeepass to add attributes marked as protected in an entry, for when I'm opening the database later with KeePass or KeePassXC.

Evidlo:
All unprotected entries are protected when the database is saved.

Evidlo:
What effect are you expecting Protected=True to have?

jeandeaual:
Sorry for the delay. Here is what I'm expecting.

When creating an entry with the following property:

```
entry.set_custom_property('Key', 'Unprotected value')
```

It will look like the following when opened with KeePassXC:

keepassxc_protect_1

When creating an entry with the following property:

```
entry.set_custom_property('Key', 'Protected value', protect=True)
```

It will look like this when opened with KeePassXC:

keepassxc_protect_2

Akianonymus:
This should be merged.

pykeepass/entry.py
```diff
@@ -81,11 +81,14 @@ def _get_string_field(self, key):
...
             return field.text
 
-    def _set_string_field(self, key, value):
+    def _set_string_field(self, key, value, protect=False):
         field = self._xpath('String/Key[text()="{}"]/..'.format(key), first=True)
         if field is not None:
             self._element.remove(field)
-        self._element.append(E.String(E.Key(key), E.Value(value)))
+        if protect:
+            self._element.append(E.String(E.Key(key), E.Value(value, Protected="False")))
```
Tejes, line 89: Why `Protected="False"` and not `True`?

jeandeaual: At the time I made this PR, https://github.com/libkeepass/pykeepass/blob/43c0d64abcbcd55e1f1e05c2e4c5d6ac0527d9ec/pykeepass/entry.py#L52 used to be
```python
E.String(E.Key('Password'), E.Value(password, protected="False"))
```
(see https://github.com/libkeepass/pykeepass/pull/186/commits/daaf7826b0eacaf0af7e867a106d5d1180fd3a81)

Also, as I mentioned it the PR description, when opening a database created by KeePass containing a protected value, it appears to have ` Protected="False"` in the raw XML.

I mimicked this behavior and it worked well, but I don't fully understand how this field works.

A6GibKm:
> What effect are you expecting Protected=True to have?

From my side I expect them to be marked as such.

Not only this allows to mark them for other clients, it also allows me to choose how to display them: If a custom attribute is protected, instead of presenting it as plain text I can present them as a password text (text replaced by bullet chars + show text button).

PR #188, opened 2020-01-11
Feature: Create Databases

rashid-nhm:
Creating databases would be a very useful feature to have for this library. This change has worked well for me and I'm sure will help out other users of this library as well. 

The logic is as follows:
 - There is a blank database added as a constant (with the password of 'password')
 - If new database needs to be created, the blank database is written to the file and the master password is changed to the user specified one.

Sample Usage:
```python

kp = create_database("/path/to/file.kdbx", password="passw0rd")

# .. use kp as a normal PyKeePass object ..
```


pykeepass/pykeepass.py
```diff
@@ -27,6 +27,36 @@
...
+QGlncf9QlHxgI5mm6D+xewwcALXYPe+twvyuHErZxgK7QIGaDp5E84d/2IJp0ZnIxo2Lqyd/C+Y9
+lMe9VdgblK2CIArcdDppe85u+nKaNlN26MEIecqYJIaZWNrAO/Boi1yANKVoSFnAywCn70nAflGT
+HPCfaczcXZEu9TrxrR+N712a4gX+ya2E2T8Uzw6fguoZj7PXhQHIKtOA3LfM7MSW5decOY1QC7Y0
+CI/9monYHCg3acQRc/EKbxsk9u5tJOh8EUw3FEYJ1Uu7RhzRun0HxPCwBgUkUdhC8ctJZXpmHD98
+sDmJOukJR3p7XbKS+9M4RWzlp4qmrdGaYqHrFMKPDO+lmmDnvaDKwvWig53XntKyqBnwt9PDia/h
+Qcy69dhZm3+kcuTZOG402QvkVe9epJE24/mip/tN3m/DrROxJs3btjqO9npNu+CdE7i6PxeUS0BX
+URGjNJsIDXvNSBlOK3xwJzY10AdNpE9MrCOnFNRAAmaZQREHetsJHkV203env5oSSn6Iv8hmjsRc
+JVqJ8WLeptZLOzGPGeOIflNqwFbzErJr0h3R7x8e7a0SUgtxx1pY7A+pqOHV7rRTF1tykKRh+iIu
+J4TfwBJP1WoF9Xiurh6DSQ99t/pFE8+gHQiU8pXIl4J5mN6tU8h5aF4pckQqkK6pINUAAAAA
+"""
```
pschmitt: I'd prefer to have this stored in an asset file. We can ship these with the pip package.

rashid-nhm: This was actually a good idea. I've moved the blank database to a separate file. This also reduced the amount of code  in the `create_database` function since there is no need to do an initial write to the user file.

pykeepass/pykeepass.py
```diff
@@ -560,3 +566,23 @@ def delete_binary(self, id):
...
+
+    with open(filename, 'wb') as f:
+        f.write(
+            KDBX.build(
+                kdbx,
+                password=password,
+                keyfile=keyfile,
+                transformed_key=transformed_key
+            )
+        )
```
Evidlo: Why aren't you using the `read` and `save` functions here?

Evidlo:
Also `create_database` needs to be added to the miscellaneous section of the readme.

pykeepass/pykeepass.py
rashid-nhm: Good point. Didn't really occur to me. I've updated the function to use the PyKeePass object instead to create the blank database.

rashid-nhm:
> Also `create_database` needs to be added to the miscellaneous section of the readme.

I have added the updated changes to the bottom of the README. You can also change it to a format that fits best.

pschmitt:
LGTM

PR #200, opened 2020-09-30
Fixes a issue with trailing bytes after xml in database files.

mikalv:
For some reason, some database files I've encountered has trailing bytes after the xml. In the last case it was b'\x06\x06\x06\x06\x06\x06' which still was read fine by most keepass applications (keepass, keeweb, etc), but which failed to load with this python library. 

The patch removes any trailing bytes from the xml before attempting to parse the xml.


Evidlo:
Please add a test into the regression tests section.

Evidlo:
Any updates on this?

Evidlo:
This was fixed in https://github.com/libkeepass/pykeepass/pull/212

PR #201, opened 2020-10-04
Add option to use streams

KeyWeeUsr:
Not in all cases is useful to use files, sometimes there's no permission to write on a specific location or simply the content needs to be kept in a variable for further processing.

This PR allows reading, writing and creating KDBX as raw bytes or reading, writing and creating KDBX into a stream (file stream in tests).

All tests passed locally for Python 3.8 + new tests added for the new functionality.


Evidlo:
How about we simply check if `filename` has a read attribute to handle streams?

Reading

    if hasattr(filename, 'read'):
        self.kdbx = KDBX.parse_stream(...)
    else:
        self.kdbx = KDBX.parse_file(...)

Writing

    if hasattr(filename, 'write'):
        KDBX.build_stream(...)
    else:
        KDBX.build_file(...)
        
If somebody wants to read in from bytes, then they should use `io.BytesIO`.

KeyWeeUsr:
@Evidlo `hasattr()` seems fine to me, but for the tests, there's missing `save()` functionality in that `test_open_save()` unlike in my tests. Also, why keep everything in a single almost 1000 lines file if you can just call `python -m unittest discover`?

Evidlo:
I modified `test_open_save()` to be more conducive to the changes you're making.

As for reorganizing the tests, that should probably go in a separate PR/issue.  Keep this PR about adding stream support.

Evidlo:
With the changes to `test_open_save()`, I think you would do something like:

``` python
stream_in = open('test3.kdbx')
stream_out = BytesIO()

filenames_in = [
    ...,
    stream_in,        # KDBX stream open test
]
filenames_out = [
    ...,
    stream_out,
]
passwords = [
    ...,
    'password',
]
...
```

PR #203, opened 2020-10-04
Move package constants from setup() into package itself

### LINKED ISSUE #199: Not possible to find out the release version
Hi there, 3.2.1 split the CredentialsIntegrity Exception into three new ones. However, when calling pykeepass it is not possible to find out which version I have loaded, so I don't know which Exceptions I can expect to use. Would it be possible to include a "pykeepass.__version__" attribute or somesuch? That would help.

KeyWeeUsr:
Move constants into the package so they can be accessed programmatically. Prevents also accidental running of `setup()` if imported (e.g. by a test) and uses a context manager to always correctly close README file when reading its contents.

Closes #199


Evidlo:
Thanks for the contribution.

I'd like to limit this to just the version.  I don't think many people will care to access 'description' and etc programmatically.

For the version, I'd prefer if you would copy the method I used here: https://github.com/Evidlo/passhole/blob/master/setup.py

pschmitt:
Can't say that I like this one either.
+1 for only doing this for the `__version__` thought.
@Evidlo why a dedicated `version.py` and not use `__init__.py` for this?

pschmitt (changes requested):
See comments on the main thread.

KeyWeeUsr:
Changed only to `__version__`.

Evidlo:
I was following recommendations [here](https://packaging.python.org/guides/single-sourcing-package-version/), but it seems like importing from `version.py` will still cause `__init__.py` to be imported anyway.  The only reason it works in passhole is because my `__init__.py` is empty.

KeyWeeUsr:
To avoid the importing we'd need to clean up `__init__` and keep only constants there. This way `__init__` in itself won't import any package which might cause importing 3rd-party deps.

Or:

```python
# __init__.py
CONSTANT = 123

try:
    from pykeepass.module import stuff
except ImportError:
    pass/if <env var>/etc
```

KeyWeeUsr:
Is there anything else needed to merge this PR? If you want I can implement the environment variable set in `setup.py` and make `__init__.py` check for it to ignore all `ImportError` exceptions when only trying to pull the constants from `version.py` file (as the first import). Or I can move it directly to `__init__.py` as well.

pschmitt:
As mentioned previously I prefer hosting the version info in `__init__.py` rather than a separate `version.py` file that only serves this one single purpose. I have to add that I'm not exactly properly educated in this regard though. What's the "pythonic" way of doing this? There's gotta be a PEP for this.

KeyWeeUsr:
According to the [`PEP-396`](https://www.python.org/dev/peps/pep-0396/#id24) it's a *module* attribute `__version__` and a module means `__init__.py`. Updated.

pschmitt (approved):
LGTM

Evidlo:
This PR seems to have broken pykeepass installation.

The line 

    from pykeepass import __version__

causes everything in `__init__.py` to be evaluated, including the imports from `pykeepass.py`.  Again python evaluates all of `pykeepass.py` and of course this fails because the dependencies are not installed yet.

See my traceback from a fresh virtualenv:

```
[evan@blackbox pykeepass] pip install -e .
Obtaining file:///tmp/pykeepass
    ERROR: Command errored out with exit status 1:
     command: /tmp/venv3/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pykeepass/setup.py'"'"'; __file__='"'"'/tmp/pykeepass/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-tyes748m
         cwd: /tmp/pykeepass/
    Complete output (9 lines):
      File "<string>", line 1, in <module>
      File "/tmp/pykeepass/setup.py", line 2, in <module>
        from pykeepass import __version__
      File "/tmp/pykeepass/pykeepass/__init__.py", line 2, in <module>
        from .pykeepass import PyKeePass, create_database
      File "/tmp/pykeepass/pykeepass/pykeepass.py", line 5, in <module>
        from future.utils import python_2_unicode_compatible
    ModuleNotFoundError: No module named 'future'
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

pschmitt:
Damn, I should have tested this. I genuinely thought this would be an easy thing to do. My bad.

Evidlo:
This reminds me that we should probably reenable automated testing on PRs.  Not sure how it got turned off in the first place.

I don't think the current Travis install command `pip install -r requirements` in `.travis.yml` would have caught this error, so we might want to adjust that.

pschmitt:
> we should probably reenable automated testing on PRs

Yeah absolutely. Been thinking about migrating to Github Actions as well.

KeyWeeUsr:
@pschmitt Ref #213 and https://github.com/libkeepass/pykeepass/pull/203#issuecomment-703905273, it'll always do that until `__init__.py` is empty because that file is loaded first as a "module" and afterwards anything else is even though you do `from pykeepass import __version__` so to do `setup.py` it'd be better to set an env variable or add `try/except` just for the case of importing.

Example:

```python
# setup.py
from os import environ
environ["PYKEEPASS_INSTALL"] = "1"
...

# __init__.py
from os import environ
if not environ.get("PYKEEPASS_INSTALL"):
    from __future__ import absolute_import
    from .pykeepass import PyKeePass, create_database
```

or a similar thing. Perhaps there's even something `setuptools` or `distutils` set in the background to the environment, so we could check for that.

PR #204, opened 2020-10-05
[WIP] use elementpath for finding via XPath (2.0)

### LINKED ISSUE #123: _xpath function throws exception when title or username contains quote
When i use function
```
kdb.add_entry(parent_kdb_group, entry.title, entry.username, entry.password, entry.url, entry.notes)
```
 and entry.title contains quote, in PyKeePass._xpath function xpath_str formed like this:
```
xpath_str = './Entry/String/Key[text()="Title"]/../Value[text()="test "inner""]/../../String/Key[text()="UserName"]/../Value[text()="some username"]/../..'
```
when tree.xpath() is runned it thows exception XpathEvalError: Invalid predicate
It happens  because syntaxis like 
>Value[text()="test "inner""] 

can't be used in xpath.

Can you please fix this problem?

KeyWeeUsr:
The problem is that [`lxml` uses XPath 1.0](https://stackoverflow.com/a/53195682/5994041). That version does not know [how to escape](https://stackoverflow.com/a/12406443/5994041) those characters as those are [special/forbidden](https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references#Predefined_entities_in_XML). Instead, we can use [`elementpath`](https://github.com/sissaschool/elementpath) which seems to be quite popular (among other XPath 2.0 related packages), but it breaks the tests.

For XPath 1.0 fixing #123 doesn't seem possible, so perhaps rewriting everything to `elementpath.select(root, path)` would work if the rest of the codebase was adjusted as well.

Closes #123

### DIFF
--- a/pykeepass/pykeepass.py
+++ b/pykeepass/pykeepass.py
@@ -10,6 +10,7 @@ import os
 import re
 import uuid
 import zlib
+import elementpath
 from copy import deepcopy
 
 from construct import Container, ChecksumError
@@ -245,8 +246,8 @@ class PyKeePass(object):
         if tree is None:
             tree = self.tree
         logger.debug(xpath_str)
-        elements = tree.xpath(
-            xpath_str, namespaces={'re': 'http://exslt.org/regular-expressions'}
+        elements = elementpath.select(
+            tree, xpath_str, namespaces={'re': 'http://exslt.org/regular-expressions'}
         )
 
         res = []
@@ -269,6 +270,12 @@ class PyKeePass(object):
             res = res[0] if res else None
 
         return res
+
+    @staticmethod
+    def _escape_xpath_quotes(value):
+        if isinstance(value, str):
+            value = value.replace("'", "''").replace('"', '""')
+        return value
 
     def _find(self, prefix, keys_xp, path=None, tree=None, first=False,
               history=False, regex=False, flags=None, **kwargs):
@@ -303,7 +310,9 @@ class PyKeePass(object):
             # handle searching custom string fields
             if 'string' in kwargs.keys():
                 for key, value in kwargs['string'].items():
-                    xp += keys_xp[regex]['string'].format(key, value, flags=flags)
+                    xp += keys_xp[regex]['string'].format(
+                        key, self._escape_xpath_quotes(value), flags=flags
+                    )
 
                 kwargs.pop('string')
 
@@ -320,7 +329,9 @@ class PyKeePass(object):
                 if key not in keys_xp[regex].keys():
                     raise TypeError('Invalid keyword argument "{}"'.format(key))
 
-                xp += keys_xp[regex][key].format(value, flags=flags)
+                xp += keys_xp[regex][key].format(
+                    self._escape_xpath_quotes(value), flags=flags
+                )
 
         res = self._xpath(
             xp,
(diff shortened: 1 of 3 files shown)


KeyWeeUsr:
Unfortunately the group lookup returns an empty list from the `elementpath.select()` and I have no clue why, so only for this single particular XPath I've changed it back to the `tree.xpath()` call and the rest works nicely with XPath 2.0. All tests passed, so the quotes + the rest works.

```
.......................................................................................
----------------------------------------------------------------------
Ran 87 tests in 12.947s

OK
```

Evidlo:
How about this function (taken from [here](https://stackoverflow.com/questions/32759318/how-to-escape-single-quote-in-xpath-1-0-in-selenium-for-python):

``` python
def escape_string_for_xpath(s):
    if '"' in s and "'" in s:
        return 'concat(%s)' % ", '\"',".join('"%s"' % x for x in s.split('"'))
    elif '"' in s:
        return "'%s'" % s
    return '"%s"' % s
```

Then we'd modify the strings in `xpath.py` and remove the double quotes around the {} substitutions and let the function above figure out quoting.

e.g.

``` python
>>> kp = PyKeePass('test4.kdbx', 'password', 'test4.key')
>>> title_xpath = '/String/Key[text()="Title"]/../Value[text()={}]/../..'
>>> title = 'quote test -> " <-'
>>> kp._xpath('//Entry' + title_xpath.format(escape_string_for_xpath(title)), cast=True, first=True)
Entry: "quote test -> " <- (None)"
```

Evidlo:
However, elementpath has no dependencies so it might be a cleaner solution if we can eliminate the ancestor hack.

Evidlo:
Is there a reason we should prefer elementpath to the above solution?  For the sake of people packaging pykeepass, I want to avoid adding any more dependencies unless its really necessary.

Evidlo:
Bump.

KeyWeeUsr:
One of the reason is that elementpath allows usage of XPath 1.0 as well as XPath 2.0 and by using the latter we're leveraging out-of-box support for quotes [and other fixes](https://stackoverflow.com/a/51377660/5994041), so overall it'd be better to use already newer model which will be even easier for switching to 3.x if/when required.

The other reason is that basically the escaping function is error prone and if there's any other character that XPath 1.0 doesn't like it'll break. Creating a list of unsupported characters and trying to escape them this way is rather ugly.

Re the dependencies reason:
* we can't just be without deps nowadays, CPython itself has quite a lot of them already just for working and lxml is one of them
* if we don't want to handle XML parsing ourselves, we need something/someone whose primary use-case is actually doing that
* elementpath has no deps currently and is pure Python

pschmitt:
Big +1 for merging this

Evidlo:
> The other reason is that basically the escaping function is error prone and if there's any other character that XPath 1.0 doesn't like it'll break. Creating a list of unsupported characters and trying to escape them this way is rather ugly.

If there are other characters besides " which are problematic, then I agree we shouldn't try to handle this ourselves.  However, I'm not aware of any others characters that have this problem yet.  Also it seems like this PR still needs special quote escaping.

As for all the new features of XPath 2.0, I don't see a need right now, as our queries in `xpath.py` are pretty simple.

In my opinion, we should keep this PR on the backburner until new issues arise, then take a second look.

PR #206, opened 2020-10-11
Add badges for IRC channel and RTD build status

KeyWeeUsr:
For IRC it also gives an online client provided directly by Freenode so there's no need to install anything (same as with Matrix).


pschmitt:
:+1:

PR #207, opened 2020-10-12
Basic CLI

### LINKED ISSUE #205: unescaped slashes for entry names/paths
Hey there,

thank you for creating this awesome library!

I have noticed that entry names are not escaped. This leads to wrong path names, as you can see here:

Bildschirmfoto 2020-10-09 um 14 37 06

Bildschirmfoto 2020-10-09 um 14 38 03

Test/Entry is a password in Test/datenbank within the path ../Datenbanken/

Is it possible to escape those names? Or any known workarounds for this?

Thanks in advance!

**EDIT**

```
database = PyKeePass("", "") 

for group in database.groups:
    print(group.name)
    print("----------------")
    for entry in group.entries:
        print(entry.path)
```

KeyWeeUsr:
Issue #205 made me thinking about the current usage and basically we have the API/backend ready, but there's an opportunity to take with missing frontend. For that we can use `argparse` and create a basic set of commands for managing kdbx from CLI. This will bring wider audience to the library especially if we'd create a PPA for Debian-based distros and provide a binary for Windows + MacOS.

It's not meant as a full standalone KeePass client as with other GUIs (although if we find a client to auto-type, it might work that way too), but as a quick tool for various bulk automation or for migration between password managers without knowing Python itself (e.g. just use bash).

Example:

```shell
$ pykeepass -p password -f ~/Desktop/mydatabase.kdbx tree
[Root]
|-- [foobar_group]
|   |-- [subgroup/name]
|   |   |-- subentry
|   |   |-- subentry2
|   |   `-- foobar_entry
|   |
|   |-- group_entry
|   `-- foobar_entry
|
|-- [foobar_group2]
|-- [Работа]
|   `-- Тест
|
|-- root_entry
`-- test/name
```

Closes #205, at least I think it'd be sufficient based on [this comment](https://github.com/libkeepass/pykeepass/issues/205#issuecomment-706529282).

### DIFF
--- a/doc/source/__main__.rst
+++ b/doc/source/__main__.rst
@@ -1,0 +1,5 @@
+__main__
+========
+
+.. automodule:: pykeepass.__main__
+   :members:
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -14,6 +14,7 @@ PyKeePass
    exceptions
    baseelement
    kdbx_parsing/*
+   __main__
 
 .. image:: https://travis-ci.org/libkeepass/pykeepass.svg?branch=master
    :target: https://travis-ci.org/libkeepass/pykeepass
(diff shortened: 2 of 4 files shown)


Evidlo:
So we talked about this a bit in the Matrix room.  I think this makes more sense for it to live in its own package/repo.

Also I mentioned the existence of [passhole](http://github.com/evidlo/passhole).  Some commands don't explicitly support non-interactive mode, but you can easily get that by piping text.

(on master branch)

``` python
entry=test_entry
username=user
url=http://example.com
password=hello

ph add ${entry} << EOF
${username}
${password}
${password}
${url}
EOF

# alternatively

ph add ${entry} <<- EOF
	${username}
	${password}
	${password}
	${url}
EOF

```

Evidlo:
Another alternative

``` bash
filename=test4.kdbx
password=password
keyfile=test4.key

entry_title=test
entry_user=user
entry_pass=pass

python - << EOF
# see pykeepass_cache if doing multiple operations
from pykeepass import PyKeePass
kp = PyKeePass("${filename}", "${password}", "${keyfile}")
print(kp.entries)
EOF
```

Thanks for the work on this nonetheless.

PR #209, opened 2020-11-17
Fix char filtering to work with narrow python

MoisMoshev:
Make it work for narrow python builds


Evidlo:
I somehow managed to mess up this PR.  Moved to https://github.com/libkeepass/pykeepass/pull/227

PR #212, opened 2020-12-05
Fix "extra content at the end of the document" error

Kaciras:
Some database file has extra content at the end of the document which cause `etree.parse` failed.

Such a file can be created by following steps:

1. Open KeePassXC 2.6.2, create a database with title 'Passwords' and password 'passwords'.

2. In Database -> Database Settings, uncheck `Enable Compression`.

3. Save Database.

The patch removes these bytes before parse.


pykeepass/kdbx_parsing/common.py
```diff
@@ -166,6 +166,8 @@ class XML(Adapter):
 
     def _decode(self, data, con, path):
         parser = etree.XMLParser(remove_blank_text=True)
+        # Remove extra content at the end
+        data = data[:data.rfind(b'>') + 1]
```
Evidlo, line 170: This might work, but I think the correct fix is to unpad inside the `DecryptedPayload` class.

``` python
class DecryptedPayload(Adapter):
    """Encrypted Bytes <---> Decrypted Bytes"""

    def _decode(self, payload_data, con, path):
        cipher = self.get_cipher(
            con.master_key,
            con._.header.value.dynamic_header.encryption_iv.data
        )
        payload_data = cipher.decrypt(payload_data)
        payload_data = CryptoPadding.unpad(payload_data, 16)

        return payload_data
...
```

tests/tests.py
```diff
@@ -801,7 +801,8 @@ def test_open_save(self):
             os.path.join(base_dir, 'test4_hex.kdbx'),       # legacy 64 byte hexadecimal keyfile test
             os.path.join(base_dir, 'test3.kdbx'),           # KDBX v3 transformed_key open test
             os.path.join(base_dir, 'test4_hex.kdbx'),       # KDBX v4 transformed_key open test
-            stream
+            stream,
+            os.path.join(base_dir, 'extra_content.kdbx')    # has trailing bytes after the xml
```
Evidlo, line 805: Maybe a more accurate name is `test4_uncompressed.kdbx`?

tests/tests.py
```diff
@@ -849,7 +853,8 @@ def test_open_save(self):
             'test4_hex.key',
             None,
             None,
-            'test3.key'
+            'test3.key',
```
Evidlo, line 856: Since this is a kdbx4 database, use `test4.key`.

tests/tests.py
Kaciras: In my test, not all uncompressed database have extra content.

pschmitt:
`test4_aes_uncompressed.kdbx` is missing. Does anyone have this file on disk?

Evidlo:
@pschmitt Added

PR #215, opened 2020-12-08
🚸 Switch to GitHub Actions and pyproject.toml

### LINKED ISSUE #214: Drop Python2 support
We really should drop Python 2.7 support. It's dead already.

https://www.python.org/doc/sunset-python-2/

pschmitt:
# Changes
- Add dependabot config
- Switch to GitHub Actions (RIP Travis CI)
- flake8 checks
- Run black and isort on the entire codebase
- **Drop Python 2 support** (Fix #214)
- Add pyproject.toml (poetry)

# TODO
- [x] flake8 checks
- [x] Fix tests (see https://github.com/libkeepass/pykeepass/pull/212#issuecomment-740475137 and [CI build logs](https://github.com/pschmitt/pykeepass/runs/1516763425?check_suite_focus=true)) [Done](https://github.com/pschmitt/pykeepass/actions/runs/410129975)
- [x] Remove `setup.py` and `requirements.txt` -> `pyproject.toml`
- [ ] Remove `requirements-rtd.txt`. Is this file even used anywhere?
- [ ] Publish to pypi using Github Actions CI
- [ ] Bump version to 4.0? (breaking change: no more Py2k)


pykeepass/pykeepass.py
```diff
@@ -106,7 +106,7 @@ def read(self, filename=None, password=None, keyfile=None, transformed_key=None)
                     transformed_key=transformed_key,
                 )
 
-        except pkpexc.ChecksumError as e:
+        except ChecksumError as e:
```
Evidlo, line 109: We need to add a test for this.

Evidlo:
It seems like you put a lot of work into this, so I feel sort of bad for even asking, but can we stick with pip and travis?  

Those are the tools I know the best.  I'm not sure we need the dependency management features of Poetry and the Github CI config looks complicated compared to travis.yml.

If you're fine with that, I can take care of it so you don't have to do all that work again.

jugmac00:
Hi,

I hope it is ok to comment here, although I am not a maintainer here.

I use libkeepass for my cli application https://github.com/jugmac00/hibpcli which combines keepass and with the have I been pwned API.

Thanks so much for your library!

Back to topic:
Similar to @Evidlo I am super familiar with Travis, and I like it a lot, it is super easy to configure, at least compared to GitHub actions, but since the company behind Travis was bought, it is getting worse and worse.. At first you needed to apply for free credits for open source projects, and now, this is even on hold:
https://twitter.com/jugmac00/status/1336239667281350656
I know quite some open source projects where CI currently is stalled.

The most open source projects I know and contribute to, go with GitHub actions. There aren't too many alternatives :-/

I won't comment on poetry/packaging - there are just too many options and opinions, but just in case you have not heard of...

There is a "linter runner" called https://pre-commit.com/ - which very comfortable downloads linters (e.g. black, flake8, ...), pins the version, runs them, and so on...

Evidlo:
Thanks, I didn't know about Travis.

Another reason to stay with pip is that there are packagers (e.g. OpenSUSE, Debian, Arch) that use build systems that pull in setup.py and its dependencies for semi-automated packaging.

pschmitt:
1. Travis vs Github Actions

The Travis build has been broken for quite some time and no one even noticed. Aside from the technical advantages, switching to GitHub Actions will give the builds better visibility. On a sidenote I honestly don't see Travis CI surviving much longer given what GitHub Actions has become capable of doing in the last few months. We are hosting pykeepass on GitHub - why not take advantage of all the tooling GitHub has to offer?

2. Removing the legacy setup.py

This will not be an issue. pip handles pyproject.toml files just fine. You pip install this branch, you'll see :)
I've done a fair amount of packaging myself and I have never seen any package pipeline that's not using pip.

Evidlo:
I just toggled pykeepass on/off in the Travis interface and it's working again.

pschmitt:
> I just toggled pykeepass on/off in the Travis interface and it's working again.

Yeah that's future-proof AF, @Evidlo

Evidlo:
Can we at least split this into separate PRs (i.e. pyproject.toml migration, flake8/Black cleanup, Github CI migration, dropping Python 2) so these changes can be trialed separately?

pschmitt:
I was just about to contact you directly in this regard.

I was aiming for a big 4.0 in one go and dev from here. But I get your point. It does make sense. I will break this up over the next few days.

Are you on board for the github action part at least?

pschmitt:
See #217 and #218 for the GitHub CI part

Evidlo:
Some potential problems with switching to pyproject.toml

1. We need to figure out how we want to single source the package version.

2. Editable installs with pip still require a placeholder setup.py

https://twitter.com/pganssle/status/1241161328137515008

pschmitt:
> 1. We need to figure out how we want to single source the package version.

I would rather have version be defined both in `pyproject.toml` and `pykeepass/version.py` and then add a CI action that checks if both are in sync.

> 2. Editable installs with pip still require a placeholder setup.py

I don't use this particular feature but judging by the twitter thread you mentioned we could get away with a setup.py like the following:

```python
#!/usr/bin/env python

import setuptools

if __name__ == "__main__":
    setuptools.setup()
```

I'll give it a try.

pschmitt:
Update: I got `pip install -e .` to work by adding the stub `setup.py` and adding `setuptools` to `build-system.requires` in `pyproject.toml`

Issue #219, opened 2020-12-15
[Security vulnerability] Seeds are not changed on database save


basak:
Hi,

Thank you for pykeepass!

I noticed what I believe is a security with the way that pykeepass saves a database (I've tested only KDBX). None of the four seeds are regenerated. These are:
 * `kdbx.header.value.dynamic_header.master_seed.data`
 * `kdbx.header.value.dynamic_header.encryption_iv.data`
 * `kdbx.header.value.dynamic_header.transform_seed.data`
 * `kdbx.header.value.dynamic_header.protected_stream_key.data`

I believe this results in a number of issues. Note that I'm not a cryptographic expert. However, the official keepass client does in my testing regenerate these parameters on every save, so I think that should be a good enough reason for pykeepass not to vary in behaviour from that.

Some issues I believe exist with not regenerating the seeds:
 * Everyone who uses `create_database()` will get the same seed for password derivation, which makes them vulnerable to precomputation attacks.
 * I'm looking to use pykeepass to derive a "subset" password database for lower security devices. In this case, although I might generate the derived database from a master database, I would not expect the derived database to leak any information about the master database I specifically chose not to include. However, without regeneration of the seeds, an attacker with access to the derived database would be able to precompute tables that make the master password faster to crack if it were later exposed.
 * Not regenerating `protected_stream_key` is not as severe, but it would still allow an attacker who manages to access only the decrypted XML to be able to compare passwords from different users for equality, when regenerating this header field would have prevented it.

Note that this affects both `create_database()` and `PyKeePass.save()`. In both cases, all seeds should be regenerated, just like the "official" keepass client does it.

Here are test cases that should pass when this issues is fixed:

    import pykeepass
    import pytest

    EXPECTED_SALT_ATTRS = [
        "kdbx.header.value.dynamic_header.master_seed.data",
        "kdbx.header.value.dynamic_header.encryption_iv.data",
        "kdbx.header.value.dynamic_header.transform_seed.data",
        "kdbx.header.value.dynamic_header.protected_stream_key.data",
    ]

    def get_recursive_attr(obj, name):
        try:
            idx = name.index(".")
        except ValueError:
            return getattr(obj, name)
        else:
            return get_recursive_attr(getattr(obj, name[:idx]), name[idx + 1 :])

    @pytest.mark.parametrize(["attr"], [(attr,) for attr in EXPECTED_SALT_ATTRS])
    def test_creation_salts_differ(tmpdir, attr):
[...]

PR #222, opened 2020-12-27
kdbx3: fix length of transform_rounds field

basak:
The official KeePass client produces and expects a 64 bit field, not a
32 bit field, for this parameter.

Using the incorrect length happens to parse just fine. The correct
length is encoded into the header so there is no parsing misalignment.
And because the field is little endian, parsing the field itself as a 32
bit field even though it is actually a 64 bit field also works, provided
that the field value fits into 32 bits, which it typically does.

However, if we write header field back out, we write it with a 32 bit
length, and KeePass rejects this. We weren't writing out the header
ourselves previously because kdbx.header.data wasn't being removed (see
https://github.com/libkeepass/pykeepass/issues/219#issuecomment-748692045).
However if we start doing that, such as to fix issue #219, then this bug
is revealed.

The fix is trivial: we change the declaration to be a 64 bit field to
match the official client.

### DIFF
--- a/pykeepass/kdbx_parsing/kdbx3.py
+++ b/pykeepass/kdbx_parsing/kdbx3.py
@@ -3,8 +3,8 @@
 
 import hashlib
 from construct import (
-    Byte, Bytes, Int16ul, Int32ul, RepeatUntil, GreedyBytes, Struct, this,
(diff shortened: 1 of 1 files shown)


Evidlo:
Thanks for finding this.

PR #227, opened 2021-01-05
Fix char filtering to work with narrow python

Evidlo:
Continuation from https://github.com/libkeepass/pykeepass/pull/209

### DIFF
--- a/pykeepass/kdbx_parsing/common.py
+++ b/pykeepass/kdbx_parsing/common.py
@@ -171,6 +171,15 @@ class XML(Adapter):
         return etree.tostring(tree)
 
 
+# FIXME python2
+def valid_xml_char(ch):
+    ch_code = ord(ch)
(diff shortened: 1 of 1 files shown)


Evidlo:
The alternative Unicode stripping method seems to be pretty slow, probably because we are iterating every character in the XML.  I added a commit which only falls back to this method if we are on a narrow build.

Can you confirm that the tests pass on your narrow 2.7 build on Windows?

Evidlo:
@MoisMoshev bump

MoisMoshev:
Hey, I can confirm that the tests pass on Python 2.7. As already said though, since 2.7 is past EOL, I do not insist on this being merged.

Evidlo:
We've sinced dropped Python 2 support, so I'm closing this.

MoisMoshev:
Yay, we are finally done with Py 2!

PR #229, opened 2021-01-08
Save to temporary file

Evidlo:
Save to `database.kdbx.tmp`, then move over `database.kdbx` to prevent database clobbering if there is an exception.

### DIFF
--- a/pykeepass/pykeepass.py
+++ b/pykeepass/pykeepass.py
@@ -8,7 +8,7 @@ import base64
 import logging
 import os
 import re
-import shutil
+import tempfile
 import uuid
 import zlib
 from copy import deepcopy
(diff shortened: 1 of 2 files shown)


pykeepass/pykeepass.py
```diff
@@ -132,27 +133,34 @@ def save(self, filename=None, transformed_key=None):
...
-                filename,
-                password=self.password,
-                keyfile=self.keyfile,
-                transformed_key=transformed_key
-            )
-        return output
+            # save to temporary file to prevent database clobbering
+            # see issues 223, 101
+            # FIXME python2 - use pathlib.Path.withsuffix
+            filename_tmp = filename + '.tmp'
```
pschmitt, line 152: Wouldn't it make more sense to use the [`tempfile` module](https://docs.python.org/3/library/tempfile.html) here?

Evidlo: Fixed

Evidlo:
this has been merged

PR #230, opened 2021-01-08
Save to BytesIO before writing to file

Evidlo:
Prevent database clobbering by saving to stream in memory before overwriting database.


A6GibKm:
There is https://github.com/rec/safer for safer `write()` operations, see https://stackoverflow.com/questions/2333872/how-to-make-file-creation-an-atomic-operation.

PR #231, opened 2021-01-09
Add BaseElement.kp getter

hxss:
Its allow to pass only element as main dependency

### DIFF
--- a/pykeepass/baseelement.py
+++ b/pykeepass/baseelement.py
@@ -56,6 +56,10 @@ class BaseElement(object):
         self._element.append(getattr(E, tag)(value))
 
     @property
+    def kp(self):
+        return self._kp
+
+    @property
     def group(self):
         return self._xpath(
             '(ancestor::Group)[last()]',


Evidlo:
`BaseElement._kp` is already exposed though?

hxss:
Yes, I'm talking about access to kp object from BaseElement instance. Isn't `_` prefix means protected property that shouldn't be used outside from the class?

Evidlo:
Yes, but that's just by convention. Python doesn't really have such a thing as protected attributes.  Things starting with `__` are another story though.

Why do you need `_kp` though?  Why not just pass the PyKeePass object along as well and make your code more explicit?

hxss:
I'm write [keepass navigation class](https://gitlab.com/hxss-linux/keeprofi/-/blob/master/keeprofi/state/kp_navigation.py#L22) that works with `group`, but its uses `_kp.filename` property in one single case(line 40)

Evidlo:
I'd say either pass in the filename or database object directly.

Evidlo:
Also, are you aware of [keepmenu](https://github.com/firecat53/keepmenu)?

hxss:
yeah, I saw keepmenu.
The problem is that I should spin extra dependency between states for only one case in one state while my main dependency(`BaseElement`) already has necessary object link. Also `Entry` object allows to get its parent by public `group` getter but `PyKeePass` object is protected for some reason.

Evidlo:
Anyway, `_kp` is likely not going anywhere any time soon, so use that if you wish.

PR #232, opened 2021-01-16
Allow using pycryptodome alternatively

iSchluff:
Hi,

As of late gentoo has deprecated pycryptodomex, because they refuse to keep the same package around twice (https://github.com/gentoo/gentoo/commit/58d1a1ddba4d5834337df0fe684d20792709d39c#diff-e44de67df9ef089ce413d3b684a23f98d1631e3202ee7512b769d8e6a55be76a). It would be cool if pykeepass could use either of them as dependency.

This commit changes the import to try load pycryptodomex first and fall back to pycryptodome.

Greetings, Anton

### DIFF
--- a/pykeepass/kdbx_parsing/common.py
+++ b/pykeepass/kdbx_parsing/common.py
@@ -1,6 +1,11 @@
-from Cryptodome.Cipher import AES, ChaCha20, Salsa20
+# support both pycryptodomex (Cryptodome) and pycryptodome (Crypto)
+try:
+    from Cryptodome.Cipher import AES, ChaCha20, Salsa20
+    from Cryptodome.Util import Padding as CryptoPadding
+except ImportError:
+    from Crypto.Cipher import AES, ChaCha20, Salsa20
+    from Crypto.Util import Padding as CryptoPadding
 from .twofish import Twofish
-from Cryptodome.Util import Padding as CryptoPadding
 import hashlib
 from construct import (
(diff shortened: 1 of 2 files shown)


pykeepass/kdbx_parsing/common.py
```diff
@@ -1,6 +1,10 @@
-from Cryptodome.Cipher import AES, ChaCha20, Salsa20
+try:
```
pschmitt, line 1: Can you please add a comment explaining why this is done? For future ref.

Evidlo:
A situation we'd like to avoid is accidentally importing the original pycrypto, which isn't being updated anymore and is missing some crypto algorithms we use.

It would be good to raise an ImportError in that case.

Evidlo:
Is there a reason why this can't be moved into a patch on the Gentoo side?

iSchluff:
Well a distro patch is only useful for the distro and has to be maintained there across updates, this patch makes it easier for all users of pykeepass to use the package and is trivial to maintain.

```python
from Crypto.Cipher import AES, ChaCha20, Salsa20
```
this line is already incompatible with pycrypto, so nothing additional to do here.

```python
>>> from Crypto.Cipher import AES, ChaCha20, Salsa20
  File "<stdin>", line 1, in <module>
ImportError: cannot import name ChaCha20
```

From my point of view the incompatibility with pycrypto should be solved through your package manager, be it pip or the distro package manager. Most distros all just carry pycryptodome/x anyway.

Evidlo:
If there are other distros that have eliminated the pycryptodomex package, then we can discuss this more.  For now it seems like a self-inflicted problem in Gentoo, so it should be solved there.

PR #261, opened 2021-05-21
Fix for setup.py, currently the dist package contains not just the pykeepass package but also the tests

sbremer:
We are experiencing some conflicts due to the pykeepass pypi-dist-package not just containing the "pykeepass" python package but also the "tests" package.
With this small fix, the dist/wheel will not contain the tests package.


pschmitt:
took me while to check it out but lgtm! thanks.

PR #279, opened 2021-10-03
implements MasterKeyChangeForce Meta key

### LINKED ISSUE #269: Support for "Force master key change on next open"
It would be a major help if the library could support setting the necessary database flags/values for "Force master key change on next open".

This would allow for the automatic generation of KeePass databases, that could then be subsequently distributed to a group of administrators - with each being required to use their own unique master key.

For my specific scenario, I do not need the library to 'honour' the flag (if already set) while opening the database, just to be able to set it for subsequent distribution.

If I can provide any further details, please ask!

Many thanks!

KeyWeeUsr:
This PR implements the forceful password change trigger which happens once a person opens the DB in a KDBX-compatible UI tool.

* The key name is defined [here](https://github.com/dlech/KeePass2.x/blob/6bb7f0cbe1afa5365fcdbed7adc124e82ac01fef/KeePassLib/Serialization/KdbxFile.cs#L118) and is always present in the DB XML.
* The type is always [C# Long](https://github.com/dlech/KeePass2.x/blob/6bb7f0cbe1afa5365fcdbed7adc124e82ac01fef/KeePassLib/Serialization/KdbxFile.Read.Streamed.cs#L248)
* The behavior is [triggered after >= 0 check](https://github.com/dlech/KeePass2.x/blob/0defb69f48687de62a4dbfad3213387371a3e8be/KeePass/Forms/DatabaseSettingsForm.cs#L235)

Closes #269

### DIFF
--- a/pykeepass/pykeepass.py
+++ b/pykeepass/pykeepass.py
@@ -743,6 +743,32 @@ class PyKeePass(object):
         for reference in binaries_gt:
             reference.id = reference.id - 1
 
+    @property
+    def needs_password_change(self):
+        """Check whether a password change is necessary."""
+        return self.password_change_force_days >= 0
+
+    @property
+    def password_change_force_days(self):
+        """Check amount of days until password force reset."""
+        path = '/KeePassFile/Meta/MasterKeyChangeForce'
+        item = self._xpath(path, first=True)
+        value = item.text
+
+        assert value  # should be always present and non-empty
+        return int(value)
+
+    @password_change_force_days.setter
+    def password_change_force_days(self, value):
+        """Set amount of days until password force reset.
+
+        Args:
+            value (`int`): Number of days until reset (-1 to disable)
+        """
+        path = '/KeePassFile/Meta/MasterKeyChangeForce'
+        item = self._xpath(path, first=True)
+        item.text = repr(value)
+
 
 def create_database(
         filename, password=None, keyfile=None, transformed_key=None


KeyWeeUsr:
Simple example:

```python

db = create_database("db.kdbx")
assert db.password_change_force_days == -1
assert not db.needs_password_change
print(db.xml())

# open with the password for blank DB (embedded in the codebase) via UI tool
# no effect, close

# enforce password change
db.password_change_force_days = 0  # immediately
assert db.password_change_force_days == 0
assert db.needs_password_change
db.save()
print(db.xml())

# open with the password for blank DB (embedded in the codebase) via UI tool
# UI needs to enforce master password change immediately
```

pschmitt:
lgtm

KeyWeeUsr:
@pschmitt Please approve it via [`Review changes`](https://github.com/libkeepass/pykeepass/pull/279/files) if it's okay or add `hacktoberfest-accepted` label for it. Thanks! :)

Evidlo:
We need to check when the masterkey was set for this to make sense.  I don't see any datetime parsing happening here, so I don't understand how this is supposed to work.

These are the fields we need to read/set:

```
    <MasterKeyChanged>2017-06-17T00:38:50Z</MasterKeyChanged>
    <MasterKeyChangeRec>-1</MasterKeyChangeRec>
    <MasterKeyChangeForce>100</MasterKeyChangeForce>
```

Evidlo:
Looking at the keepass2 code linked above, I think the booleans `bChangeRec` and `bChangeForce` are just to determine whether the checkboxes in the settings menu are ticked or not: 
out

Evidlo:
This was implemented in https://github.com/libkeepass/pykeepass/commit/9811ac0a4603db7c862b7f69fb55681505a19c59

PR #280, opened 2021-10-03
Fix kdbx nuke

### LINKED ISSUE #266: Error parsing path on a virtual filesystem
when trying to `save` on a virtual filesystem on a Flatpak environment 
I get an error of type `construct.core.StreamError` with message `Error in path (building) -> header`. The path of such file is of the form
```
/run/user/1000/gvfs/google-drive:host=domain.com,user=xxxxxx/aosifj--IDAKSK/ASNsakdsuMOS_asoiamaoSKFN
```

KeyWeeUsr:
I was trying to reproduce the #266 issue with Google Drive + Gnome but I couldn't find any relevant `seek()` that would break it within the `construct` library nor any file from the `gvfs` behaved strangely. If anything, the issue might need a better traceback because [`self.build_stream(...)`](https://github.com/construct/construct/blob/2626b398a676191114f6ce5d8db903e9590700db/construct/core.py#L355) calls further and it's still Python, so it shouldn't stop somewhere in the middle of the call stack if it's able to produce an `OSError`.

While the #266 isn't strictly a PyKeePass issue (and it's not reproducible clearly), I think leaving a free hand for the package to break the DB isn't really useful either. On top of that, some special files on Linux are actually sockets which may also happen to be used, thus it also makes sense to allow output to a socket.

```bash
# "server" / socket dumping the content
nc -lU socket.kdbx > file.kdbx
```
```python
# client
db = create_database("hello.kdbx")  # or open an existing one
db.save("socket.kdbx")
# socket then receives the bytes and dumps them to the "file.kdbx"
```

Closes #266

### DIFF
--- a/pykeepass/pykeepass.py
+++ b/pykeepass/pykeepass.py
@@ -10,8 +10,11 @@ import os
 import re
 import uuid
 import zlib
+import socket
 from copy import deepcopy
 from os.path import exists
+from stat import S_ISSOCK
+from io import BytesIO
 
 from construct import Container, ChecksumError
 from lxml import etree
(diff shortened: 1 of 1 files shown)


Evidlo:
Sorry I let this languish.  I implemented safe saving using `BytesIO`, which also means that we don't seek on the write medium.

Issue #194, opened 2020-04-05
Entry passwords are made unprotected by default


keepassium:
Hi,

It seems that pykeepass makes password strings unprotected in the generated XML:

    <String>
        <Key>Password</Key>
        <Value Protected=\"False\">super_secret_password</Value>
    </String>
 
Usually, there are good reasons to protect the passwords in memory. For instance, databases created by KeePass 2 have that attribute set to True.

However, pykeepass [specifically](https://github.com/libkeepass/pykeepass/blob/master/pykeepass/entry.py#L52) resets the protection flag:

    self._element.append(
        E.String(E.Key('Password'), E.Value(password, protected="False"))
    )

But why? This weakens data protection, and the user won't even know about that...

(Kudos to @r0bc94, who first reported this here: https://gitlab.gnome.org/World/PasswordSafe/-/issues/137)

rashid-nhm:
Currently is there a way to specify `entry.protected` in any way?

Elias481:
The title of the Bug is not very correct.
pykeepass specifially SETs the protection flag at mentioned [location](https://github.com/libkeepass/pykeepass/blob/master/pykeepass/entry.py#L52).
The existence of the protected flag means, that the value is to be protected on write - and had been unprotected on read if it was st to True before. Implemented in [class UnprotectedStream(Adapter)](https://github.com/libkeepass/pykeepass/blob/4ef5069d14131b2ac8d34ef260afdbd58bffb3a4/pykeepass/kdbx_parsing/common.py#L174).

The problem ist a bug in that class, a regression introduced with [this commit](https://github.com/libkeepass/pykeepass/commit/a6dd83d5ab9f27b1de3fef7ac5c96f5eff979bbd).
During write of KDBx, [a copy of the in-memory XML structure is made](https://github.com/libkeepass/pykeepass/blob/4ef5069d14131b2ac8d34ef260afdbd58bffb3a4/pykeepass/kdbx_parsing/common.py#L202) because the in-memory working-structure is desingend to be unprotected all the time in pykeepass and to save mens not to close. [But the return value of the function](https://github.com/libkeepass/pykeepass/blob/4ef5069d14131b2ac8d34ef260afdbd58bffb3a4/pykeepass/kdbx_parsing/common.py#L212) is then the original unencrpted in-memory version of the XML and not the version encrypted for writing out to KDBx.

[...]

PR #294, opened 2022-01-10
Allow extended properties on elements

eplouf:
This relates to https://github.com/libkeepass/pykeepass/issues/194
 This patch allows to put an entry of a custom_property to protected memory, with
 argument Protected=True.
 Could be use for future properties, as it doesn't constrain on a fixed property
 name.


Evidlo:
This was fixed in https://github.com/libkeepass/pykeepass/pull/186

PR #322, opened 2022-07-16
entry: Add is_custom_property_protected method

A6GibKm:
In Secrets we want to present protected custom properties with hidden text, so we need a way to query whether an attribute is protected.

My knowledge of lxml is just what you can see in the PR, so please tell me if there is a better way to achieve this.

See https://gitlab.gnome.org/World/secrets/-/merge_requests/721, specifically https://gitlab.gnome.org/World/secrets/-/merge_requests/721/diffs#234455d6642dc1a00aa62f6a0b9581fc403591c1_611_615 where we use this ~~hack~~ *method*.

### DIFF
--- a/pykeepass/entry.py
+++ b/pykeepass/entry.py
@@ -294,6 +294,25 @@ class Entry(BaseElement):
             raise AttributeError('Could not find property element')
         self._element.remove(prop)
 
+    def is_custom_property_protected(self, key):
+        """Whether a custom property is protected.
+
+        Return False if the entry does not have a custom property with the
+        specified key.
+
+        Args:
+            key (:obj:`str`): key of the custom property to check.
+
+        Returns:
+            bool: Whether the custom property is protected.
+
+        """
+        assert key not in reserved_keys, '{} is a reserved key'.format(key)
+        field = self._xpath('String/Key[text()="{}"]/../Value'.format(key), first=True)
+        if field is not None:
+            return field.attrib.get("Protected", "False") == "True"
+        return False
+
     @property
     def custom_properties(self):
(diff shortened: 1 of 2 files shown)


A6GibKm:
Ill add test later today

pykeepass/entry.py
```diff
@@ -294,6 +294,25 @@ def delete_custom_property(self, key):
...
+        assert key not in reserved_keys, '{} is a reserved key'.format(key)
+        field = self._xpath(
+            'String/Key[text()="{}"]/../Value'.format(key), first=True
+        )
+        if field is not None:
+            for key_, value in zip(field.keys(), field.values()):
+                if key_ == "Protected" and value == "True":
+                    return True
+
+        return False
```
Evidlo, line 314: Should be able to just do `return bool(field.attrib.get("Protected", False))

pykeepass/entry.py
```diff
@@ -294,6 +294,25 @@ def delete_custom_property(self, key):
             raise AttributeError('Could not find property element')
         self._element.remove(prop)
 
+    def is_custom_property_protected(self, key):
+        """Whether the attribute with a specific key is protected.
+
+        If the key does not exists it returns False.
+
```
Evidlo, line 301: Needs an "Args:" section

pykeepass/entry.py
A6GibKm: Thanks, note though that `bool("False")` actually returns `True` so one has to compare it with the string `"True"`.

A6GibKm:
Is the test in the correct class?

Evidlo:
Test looks good.  Thanks for the contribution!

PR #324, opened 2022-10-23
Make sure the testsuite does not leave artifacts behind

baldurmen:
Hello!

While working on the Debian package for pykeepass, I found out when the testsuite is ran, `db_dump.xml` is left behind in the CWD.

This commit makes so that it runs cleanly.

Cheers,


Evidlo:
Maybe it would be better just to use `tempfile.TemporaryFile` directly?

baldurmen:
It could be done and I certainly wouldn't be against it, but this is really a drive-by patch. It was simple enough that I submitted a patch instead of opening a bug report.

I'm not sure I have the motivation (even if it's a pretty simple change) to make extra changes though...

Bonus points: I've tested this patch and it works :)

Thank you for caring and for your work on this project though.

PR #325, opened 2022-11-13
Support for "Enable Searching" property of groups

infeeeee:
Hello

Added support for reading and setting the `EnableSearching` parameter. Other keepass clients use this to set that search should find results in that group. 

On KepassXC and Keepass the trash group has this option enabled by default, I added the same behavior as well.

I didn't modify the `find_`... functions. Should I add support there as well, or only clients should implement this search filtering?

I added some tests for this new thing, and added some generic group attribute tests based on the similar entry tests. Should I add more, or is this enough?

Edit: TODO:

- [x] Rename to `searching_enabled`
- [ ] Add support in the `find_`... functions
- [ ] Add tests for the `find_`... functions
- [ ] Update readme

### DIFF
--- a/pykeepass/pykeepass.py
+++ b/pykeepass/pykeepass.py
@@ -434,13 +434,16 @@ class PyKeePass(object):
         return res
 
     # creates a new group and all parent groups, if necessary
-    def add_group(self, destination_group, group_name, icon=None, notes=None):
+    def add_group(self, destination_group, group_name, icon=None, 
+			notes=None, enable_searching=None):
         logger.debug('Creating group {}'.format(group_name))
 
         if icon:
-            group = Group(name=group_name, icon=icon, notes=notes, kp=self)
+            group = Group(name=group_name, icon=icon, notes=notes, 
+				kp=self, enable_searching=enable_searching)
         else:
-            group = Group(name=group_name, notes=notes, kp=self)
+            group = Group(name=group_name, notes=notes, kp=self, 
+				enable_searching=enable_searching)
         destination_group.append(group)
 
         return group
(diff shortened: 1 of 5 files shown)


Evidlo:
Thanks for the contribution.

> I didn't modify the find_... functions. Should I add support there as well, or only clients should implement this search filtering?

That would definitely be desirable.  I think you only need to modify `xpath.py` to add this.  See `autotype_enabled` as an example.  Also add this to this `find_groups` section in the README.

I think `searching_enabled` would be a better name for this property so that it matches existing vocabulary like `autotype_enabled`.

pykeepass/pykeepass.py
```diff
@@ -434,13 +434,16 @@ def find_groups(self, recursive=True, path=None, group=None, **kwargs):
         return res
 
     # creates a new group and all parent groups, if necessary
-    def add_group(self, destination_group, group_name, icon=None, notes=None):
+    def add_group(self, destination_group, group_name, icon=None, 
+			notes=None, enable_searching=None):
```
Evidlo, line 438: Please use spaces instead of tabs.

infeeeee: Fiexd

Evidlo:
If you need any more direction on what to do, let me know.

infeeeee:
I didn't have time nowadays to work on this, but I still want to finish it sometimes in the future. This xpath is a bit confusing, I used lxml before on another project, but never used this part.

If you want to finish it, I won't be mad :)

infeeeee:
I stopped using keepass and switched to a different password manager, so I won't finish this PR, sorry.

PR #343, opened 2023-05-23
Use __hash__ for BaseElement

rfinnie:
- Allows for e.g. set() which uses hash() for deduplication
- Consider mtime for HistoryEntry equality
- Additional unit tests

### DIFF
--- a/pykeepass/baseelement.py
+++ b/pykeepass/baseelement.py
@@ -166,11 +166,13 @@ class BaseElement(object):
     def __repr__(self):
         return self.__str__()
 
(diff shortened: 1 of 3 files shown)


Evidlo:
Why have a separate `_hash_key` property?  Why not put this directly in `__hash__` and then in `__eq__` test for `hash(self) == hash(other)`?

rfinnie:
My key handling code was originally more complex while trying to figure out how to handle HistoryEntry, and I never questioned keeping the property after verifying uuid + mtime was accepted. PR updated.

PR #347, opened 2023-08-15
entry: move entries within a group

pwall2222:
Code changes:
- Add `index` property to Entry
- Add `move` method to Entry
- Add `_first_entry` property to Group

Explanation:
To move Entries around we need to remove and then insert at the desired index, the KDBX format makes it so that Entries and group info is all mixed, so we have to get the first element index to position correctly the Entries.

If you know a better method to get the index of the first entry I'll put it in.

What it does:

It moves entries around a group, pretty self explanatory, like the Keepass and KeepassXC functionality.

### DIFF
--- a/pykeepass/entry.py
+++ b/pykeepass/entry.py
@@ -119,6 +119,20 @@ class Entry(BaseElement):
             return results
 
     @property
+    def index(self):
+        group = self.group._element
+        children = group.getchildren()
+        first_index = self.group._first_entry
+        index = children.index(self._element)
+        return index - first_index
+
+    def move(self, new_index):
+        group = self.group._element
+        first_index = self.group._first_entry
+        group.remove(self._element)
+        group.insert(new_index+first_index, self._element)
+
+    @property
     def attachments(self):
         return self._kp.find_attachments(
             element=self,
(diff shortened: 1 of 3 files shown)


pwall2222:
Is just for feature parity with the `KeePassXC` client.

Evidlo:
Perhaps it would be better to call this `.reindex()` or something similar.  I'd like to avoid confusion with the `move_entry()` function, which is for hierarchical operations.

pwall2222:
As you sugggested I added a test and I rename the function to `reindex`.
The keepassxc custom reorder, but anyway it was a feature in the original keepass.
https://github.com/keepassxreboot/keepassxc/issues/2295
https://github.com/keepassxreboot/keepassxc/pull/4357

Evidlo:
Thanks!  This looks good so far.

Would you mind adding [docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) on `reindex` and `index` and also in README.md beneath `move_entry`?  Sorry, I should've mentioned that before.  After that it looks good to merge.

pwall2222:
Shouldn't it be near `Entry.delete_attachment` since its in the `Entry` class it self?

Evidlo:
> Shouldn't it be near Entry.delete_attachment since its in the Entry class it self?

Actually there doesn't appear to be sections for Group/Entry member functions in the README, so you can ignore that for now.

PR #348, opened 2023-08-15
Add pyproject.toml

pwall2222:
This project needs a `pyproject.toml` since in future versions of Python it will be required in order to install the package via pip.
It retains **100%** functionalty with the original `setup.py` having the dynamic version and all.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,13 +9,10 @@ license = {text = "GPL-3.0"}
 keywords = ["vault", "keepass"]
 dependencies = [
     "python-dateutil",
-    # FIXME python2 - last version to support python2
-    "construct==2.10.68",
+    "construct",
     "argon2_cffi",
     "pycryptodomex>=3.6.2",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -0,0 +1,31 @@
+[project]
+name = "pykeepass"
+readme = "README.rst"
+description = "Python library to interact with keepass databases (supports KDBX3 and KDBX4)"
+authors = [
+    { name = "Philipp Schmitt", email = "philipp@schmitt.co" }
+]
+license = {text = "GPL-3.0"}
+keywords = ["valorant", "vault", "keepass"]
```
Evidlo, line 9: What is valorant?

pwall2222: Nope, definitely have to delete that. Small mistake from using a template.

Evidlo:
Thanks!  Sorry it took me a while to get to this.

PR #354, opened 2023-08-21
chore(ci): bump actions, cache pip packages, etc.

rasa:
Test on 3.12
Allow testing on Windows and macOS
Allow easy commenting out of specific versions


rasa:
Once #351 is merged, python 3.11 and python 3.12 will pass all tests. See [here](https://github.com/rasa/pykeepass/actions/runs/5920828263).

Issue #359, opened 2023-09-18
Provide access to master seed


janbrummer:
We are currently developing YubiKey and Smartcard support integration for GNOME Secrets and this implementation should be of course compatible with other keepass application like KeepassXC: https://gitlab.gnome.org/World/secrets/-/merge_requests/757

For the challenge response method of YubiKey KeepassXC is making use of the master seed of the database file. The result in turn is added to the transform key.

Currently pykeepass does not provide access to the master seed WITHOUT unlocking the database. But as this is in this case necessary to compute the correct hash, we need to read the data first and do the unlock in a second step.

Could you please provide such an API?

janbrummer:
Adding @A6GibKm

Evidlo:
Sure, this is something I'm open to.  Probably the right way to do this is by adding a construct `If` conditional around the body and only parsing if some `decrypt` argument is True.

Evidlo:
Also I understand Secrets may access other properties in kdbx_parsing, such as what KDFs are available.  It would probably be good to turn this into a public API.

PR #360, opened 2023-09-20
Add decrypt option to decrypt body on demand

janbrummer:
Fixes: https://github.com/libkeepass/pykeepass/issues/359

### DIFF
--- a/pykeepass/pykeepass.py
+++ b/pykeepass/pykeepass.py
@@ -45,6 +45,9 @@ class PyKeePass():
             database is assumed to have no keyfile
         transformed_key (:obj:`bytes`, optional): precomputed transformed
             key.
+        decrypt (:obj:`bool`, optional): whether to decrypt XML payload.
+            Set `False` to access outer header information without decrypting
+            database.
 
     Raises:
         CredentialsError: raised when password/keyfile or transformed key
@@ -59,13 +62,14 @@ class PyKeePass():
     """
 
     def __init__(self, filename, password=None, keyfile=None,
-                 transformed_key=None):
+                 transformed_key=None, decrypt=True):
 
         self.read(
             filename=filename,
             password=password,
             keyfile=keyfile,
-            transformed_key=transformed_key
+            transformed_key=transformed_key,
+            decrypt=decrypt
         )
 
     def __enter__(self):
@@ -76,7 +80,7 @@ class PyKeePass():
         pass
 
     def read(self, filename=None, password=None, keyfile=None,
-             transformed_key=None):
+             transformed_key=None, decrypt=True):
         """
         See class docstring.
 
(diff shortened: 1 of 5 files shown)


janbrummer:
@Evidlo Like so?

Evidlo:
Looks good so far.  Just needs a test and a description added to the misc section of the README

janbrummer:
So i tried to implement the necessary parts in Secrets and discovered that i was wrong with master_seed... it's actually the kdf algorithm salt. Sorry about that, somehow mixed the output during debugging keepassxc. Nevertheless the PR main part is still valid. I'll adjust the property and add the necessary requested bits.

pykeepass/kdbx_parsing/kdbx.py
```diff
@@ -24,10 +24,12 @@ def check_signature(ctx):
             )
         )
     ),
-    "body" / Switch(
-        this.header.value.major_version,
-        {3: Body3,
-         4: Body4
-         }
+    If(this._.decrypt,
```
Evidlo, line 27: "body" should be outside `If`, I think.

pykeepass/pykeepass.py
```diff
@@ -209,6 +212,11 @@ def transformed_key(self):
         and passed to `open` for faster database opening"""
         return self.kdbx.body.transformed_key
 
+    @property
+    def master_seed(self):
+       """bytes: master seed of database"""
```
A6GibKm, line 217: Is there more that can be said about this? e.g. is there an expected size?

pykeepass/kdbx_parsing/kdbx.py
janbrummer: Code changed and this is obsolete.

pykeepass/pykeepass.py
janbrummer: It's no longer the master_seed but the salt and if i remember correctly this is not fixed. Any idea?

janbrummer:
Secrets works fine with this solution \o/ Ready for review.

pykeepass/pykeepass.py
```diff
@@ -45,6 +45,7 @@ class PyKeePass():
             database is assumed to have no keyfile
         transformed_key (:obj:`bytes`, optional): precomputed transformed
             key.
+        decrypt (:obj:`bool`, optional): whether to decrypt payload or not.
```
A6GibKm, line 48: I think this needs more explanation, its not clear what it does from an outsider perspective.

pykeepass/pykeepass.py
```diff
@@ -209,6 +215,12 @@ def transformed_key(self):
         and passed to `open` for faster database opening"""
         return self.kdbx.body.transformed_key
 
+    @property
+    def database_salt(self):
```
A6GibKm, line 219: Needs a test, if possible a small line explaining what it is too.

A6GibKm (commented):
Thanks for doing this PR!

pykeepass/pykeepass.py
janbrummer: Ok, then please help me out with a better description

pykeepass/pykeepass.py
janbrummer: Ok, will add one.

pykeepass/pykeepass.py
janbrummer: Ping :)

Evidlo: Maybe something like 

    whether to decrypt XML payload.  Set `False` to access outer header information without decrypting database.

janbrummer: Thanks for jumping in, i'll take it.

pykeepass/pykeepass.py
janbrummer: Added.

Issue #363, opened 2023-10-23
Feature Request: Add option to provide keyfile as bytes


jbrummer-vw:
Currently keyfile must be a link to a real file. Using a Yubikey challenge response in upcoming Secrets we could use the bytes directly to open our database. The current workaround is to create a new file just for the keyfile handover.

Can we change the current implementation to bytes or add a new option for bytes transfer?

jbrummer-vw:
@A6GibKm FYI

A6GibKm:
If I am not mistaken, the current documentation suggests that you should be able to pass a bytes like object instead of a path, so it could be patched to do so.

jbrummer-vw:
> If I am not mistaken, the current documentation suggests that you should be able to pass a bytes like object instead of a path, so it could be patched to do so.

Where is it mentioned in doc?

A6GibKm:
See https://github.com/libkeepass/pykeepass/pull/201/files, filename takes a stream which is also useful (One can use a pythonio.BytesIO). But yeah, no mention about it for keyfiles.

PR #364, opened 2023-10-23
Allow keyfile to be passed as bytes

janbrummer:
In order to prevent application to create temporary files, provide keyfile as bytes.

Fixes: https://github.com/libkeepass/pykeepass/issues/363

### DIFF
--- a/pykeepass/kdbx_parsing/common.py
+++ b/pykeepass/kdbx_parsing/common.py
@@ -105,6 +105,25 @@ def aes_kdf(key, rounds, key_composite):
     return hashlib.sha256(transformed_key).digest()
 
 
+def load_keyfile_composite(key: bytes):
+    try:
+        int(key, 16)
+        is_hex = True
+    except ValueError:
+        is_hex = False
+    # if the length is 32 bytes we assume it is the key
+    if len(key) == 32:
+        keyfile_composite = key
+    # if the length is 64 bytes we assume the key is hex encoded
+    elif len(key) == 64 and is_hex:
+        keyfile_composite = codecs.decode(key, 'hex')
+    # anything else may be a file to hash for the key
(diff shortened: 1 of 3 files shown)


janbrummer:
@Evidlo May i ask you for a review?

pykeepass/kdbx_parsing/common.py
```diff
@@ -116,43 +135,33 @@ def compute_key_composite(password=None, keyfile=None):
...
-                elif version.startswith('2.0'):
-                    # read keyfile data and convert to bytes
-                    keyfile_composite = bytes.fromhex(data_element.text.strip())
-                    # validate bytes against hash
-                    hash = bytes.fromhex(data_element.attrib['Hash'])
-                    hash_computed = hashlib.sha256(keyfile_composite).digest()[:4]
-                    assert hash == hash_computed, "Keyfile has invalid hash"
-        # otherwise, try to read plain keyfile
-        except (etree.XMLSyntaxError, UnicodeDecodeError):
+        if isinstance(keyfile, bytes):
```
A6GibKm, line 138: See the save method, where it checks whether the file has `hasattr(filename, "write")` maybe the same could be done with the read attribute. BytesIO has this btw.

janbrummer: In this case i'd need to change it from bytes to BytesIO. Not sure we want to enforce this in apps using pykeepass? NO strong opinion here, up to the reviewers i guess.

A6GibKm:
Hello sorry to ping you @Evidlo. Is anything missing on your end?

Evidlo:
Sorry for the late response.

I did a refactor which is a bit simpler than before, but it assumes that `keyfile` is a file-like object (such as BytesIO).

PR #371, opened 2024-01-04
Use datetime functions from python standard lib

ekuler:
The date handling functionality in Python std lib has been improved in Python 3 so we can take advantage of it and remove dependency on `dateutil` package.

This PR simplifies time zone handling for the various entry and group time values and creates time zone aware values everywhere.

~~Note that the parse function from `dateutils` is replaced by `.fromisoformat()` function from `datetime` class. However, this would raise the required Python version to 3.7. I believe this should be fine since 3.7 is quite old version and all major currently supported LTS Linux distributions are already on Python >=3.7.~~

~~If there is need to stay at Python 3.6, this can be rewritten in slightly more verbose way while still depending only on python standard library.~~

Using the `strptime` function now to stay compatible with Python 3.6 and to handle the time zone string .


ekuler:
Replaced the `fromisoformat` function with `strptime`. Turns out, prior to Python 3.11, `fromisoformat`  won't parse string produced with `isoformat` when timezone is specified. With this change, it should be compatible all the way to 3.6 again.

ekuler:
The CI now just fails on lxml build for 3.11. This will be resolved when https://github.com/libkeepass/pykeepass/pull/351 gets merged.

Evidlo:
Thanks!  It's always good to get rid of dependencies.

PR #372, opened 2024-01-09
Use modern Python3 syntax

ekuler:
Changes were suggested by Ruff linter. I've selected only the most obvious fixes to make it easier to review.


Evidlo:
Thanks!  Sorry it took so long.

Issue #311, opened 2022-06-18
Support for Yubikey / Hardware Key


MarcSchuh:
Are there any thoughts on supporting Yubikey for challenge response as second factor?

zbyso23:
I don't think so, this is library for manipulate with KeePass database, not whole App.

MarcSchuhTNG:
Sure, but the app uses the yubikey to hash the password to increase entropy. So to say: "Make a short easy to remember password strong" and hence it would be nice, if the same hashing function could be used by the lib.

zbyso23:
I still don't understanding why this functionality adding to library create for manipulating with keepass database? ...and how?

MarcSchuhTNG:
I understand your argument but it still leads to the problem, that I cannot access the database with this lib, if it is encrypted with the Yubikey being a second factor.

zbyso23:
what about using this library together with Yubikey library?
https://yubico-client.readthedocs.io/en/latest/

janbrummer:
I've managed to add Yubikey support into a hacked version of Secrets. In order to implement it in a sane way that is compatible with the KeepassXC implementation i need access to the master seed of the database prior to unlocking it. This seed is used in Yubikeys challenged-response mechanism. Could you please provide it?

Evidlo:
I'm open to adding some amount of support for this if we check for optional dependencies such as `yubico-client` at runtime.

janbrummer:
> I'm open to adding some amount of support for this if we check for optional dependencies such as `yubico-client` at runtime.

Although it is always nice having the necessary support in one lib, you are opening the gate for other crypto devices as well. I would request pkcs11 support afterwards and going further tpm support for it. IMHO this library should only support the core functions and leave the rest for the apps. Just my 2 cents.

janbrummer:
Maybe adding an example how to add Yubkikey support as an example would be enough? Based on the current changes in my PR?

Evidlo:
https://github.com/libkeepass/pykeepass/pull/373

PR #373, opened 2024-01-15
Implement framework for flexible 2FA

BryanJacobs:
This adds support for using the hmac-secret FIDO extension to contribute keying material for a KeePass 4 file.

It does this by storing an additional XML statekeeping blob in the outer ("public") header. This blob is designed to hold a variety of different authentication factors, such as passwords, key files, and Yubikey challenge-response devices.

The implementation here has support for passwords (primitively - no brute-force resistance), key files, and FIDO2 authenticators. It contains documentation about the changes to the KeePass file format, and tests covering the basic functionality.

It adds a dependency on `python-fido2` to perform actual FIDO2 operations with connected PC/SC or USB-HID authenticators.

### DIFF
--- a/pykeepass/deprecated.py
+++ b/pykeepass/deprecated.py
@@ -1,6 +1,3 @@
-#!/usr/bin/env python3
-
-
 # ---------- Find functions ---------------
 # Use find_entries()/find_groups() instead
 


.gitignore
```diff
@@ -9,3 +9,4 @@ Pipfile.lock
 *.kdbx
 *.kdbx.out
 .idea
+.venv
```
BryanJacobs, line 12: Not usnig an isolated Python environment rapidly leads to a mess. I've marked the demi-standard `.venv` directory as ignored to make it easier to keep a virtual environment inside the project folder.

multifactor_auth.rst
BryanJacobs, line 1: This file isn't just for `pykeepass`. I'm going to implement support for this in other projects next, and figured it would be good to have a reference, since the whole KeePass file format isn't centrally owned.

pykeepass/multifactor_format.rst
BryanJacobs, line 1: This file isn't just for `pykeepass`. I'm going to implement support for this in other projects next, and figured it would be good to have a reference, since the whole KeePass file format isn't centrally owned.

pykeepass/__init__.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 from .pykeepass import PyKeePass, create_database
+from .kdbx_parsing.factorinfo import FactorInfo, FactorGroup, FIDO2Factor, PasswordFactor, KeyFileFactor
 
 from .version import __version__
 
-__all__ = ["PyKeePass", "create_database", "__version__"]
+__all__ = ["PyKeePass", "create_database", "__version__", 'FactorInfo', 'FactorGroup', 'FIDO2Factor', 'KeyFileFactor']
```
BryanJacobs, line 7: As these classes are necessary to make a database *start* using multifactor auth, I figured it made sense to expose them at the top level.

pykeepass/fido2.py
BryanJacobs, line 1: This file - and only this file - contains code for talking to FIDO2 authenticators.

The two operations were perform are, of course, make-credential and get-assertion.

pykeepass/kdbx_parsing/common.py
```diff
@@ -105,7 +104,49 @@ def aes_kdf(key, rounds, key_composite):
     return hashlib.sha256(transformed_key).digest()
 
 
-def compute_key_composite(password=None, keyfile=None):
+def compute_keyfile_part_of_composite(keyfile):
```
BryanJacobs, line 107: Factored out of `compute_key_composite`

pykeepass/kdbx_parsing/common.py
```diff
@@ -173,6 +181,30 @@ def compute_master(context):
     return master_key
 
 
+def populate_custom_data(kdbx, d):
```
BryanJacobs, line 184: I found working with `construct` extremely difficul.

pykeepass/pykeepass.py
```diff
@@ -152,14 +176,19 @@ def save(self, filename=None, transformed_key=None):
         if not filename:
             filename = self.filename
 
+        if hasattr(self.kdbx.header, 'data'):
```
BryanJacobs, line 179: I noticed that if you called `save()` after changing anything in the header it didn't really save because the binary representation was already there...

pykeepass/pykeepass.py
```diff
@@ -854,10 +945,13 @@ def create_database(
     keepass_instance.filename = filename
     keepass_instance.password = password
     keepass_instance.keyfile = keyfile
+    keepass_instance.authentication_factors = authentication_factors
+    keepass_instance._factor_data = factor_data
 
-    keepass_instance.save(transformed_key)
+    keepass_instance.save(transformed_key=transformed_key)
```
BryanJacobs, line 951: I think this was bugged before? If I understand correctly it would pass `transformed_key` as the `filename`. So I guess nobody really used this function that way?

requirements.txt
```diff
@@ -1,5 +1,6 @@
-lxml==4.8.0
+lxml==5.1.0
```
BryanJacobs, line 1: Version bump to work with `fido2`. Doesn't seem to break things.

pykeepass/kdbx_parsing/factorinfo.py
BryanJacobs, line 1: This file contains most of the implementation heavy lift.

pykeepass/kdbx_parsing/factorinfo.py
```diff
@@ -0,0 +1,490 @@
...
+            raise NotImplementedError(
+                "Cannot unwrap a key part for unknown alg {} on factor type {}".format(self.key_type, factor_name)
+            )
+
+        digest = None
+        if self.group.validation_type == FACTOR_VALIDATE_HMAC_SHA512:
+            digest = hmac.new(decrypted_key, self.group.validation_in, 'SHA-512').digest()
+        else:
+            # Can't verify, we don't know how
+            pass
```
BryanJacobs, line 331: In the event it's wrong we'll just end up failing to open the database later, as the constructed composite key won't pass the header checksum.

pykeepass/kdbx_parsing/factorinfo.py
```diff
@@ -0,0 +1,490 @@
...
+                if unwrapped_part is not None:
+                    self.cached_key_part = (user_supplied_info, unwrapped_part)
+                    return unwrapped_part
+            except CredentialsError as e:
+                log.error("Factor failed: {}".format(e))
+                continue
+
+        next_challenge = random.randbytes(32)
+
+        if len(fido2_factors) > 0:
```
BryanJacobs, line 176: Rather important (and very inconvenient to implement...) feature here. We make one FIDO2 call per authenticator per group.

If there are ten Factor entries, each one of which is a FIDO2 authenticator... that means for each authenticator one would succeed and nine would fail.

If there's one authenticator plugged into the system it would not be great for the user to have to type their PIN ten times.

So instead we gather all the credentials in the group and send them in one call. The authenticator will accept the credential it created, and ignore the others.

pykeepass/fido2.py
```diff
@@ -0,0 +1,147 @@
...
+        ],
+        exclude_credentials=[
+            PublicKeyCredentialDescriptor(
+                type=PublicKeyCredentialType.PUBLIC_KEY,
+                id=credential_id
+            ) for credential_id in already_enrolled_credentials
+        ],
+        extensions={
+            "hmacCreateSecret": True,
+            "credentialProtectionPolicy": CredProtectExtension.POLICY.REQUIRED,
```
BryanJacobs, line 103: This forces the user to provide "verification" (a PIN, a fingerprint, etc - not just a touch) every time the credential is used.

This is important because the hmac-secret extension says that the keys used are silently different for UV (user verification) vs non-UV scenarios. It would not be great if the key were created with non-UV, and then later the authenticator is set to always-UV mode and so it becomes impossible to get the original key again.

pykeepass/fido2.py
```diff
@@ -0,0 +1,147 @@
...
+            PublicKeyCredentialParameters(
+                type=PublicKeyCredentialType.PUBLIC_KEY,
+                alg=ES256.ALGORITHM
+            )
+        ],
+        exclude_credentials=[
+            PublicKeyCredentialDescriptor(
+                type=PublicKeyCredentialType.PUBLIC_KEY,
+                id=credential_id
+            ) for credential_id in already_enrolled_credentials
```
BryanJacobs, line 99: Passing the credentials already in the group here prevents adding the same authenticator to the same group twice.

You can still add the same authenticator to two different groups, and of course to two different files. But I don't think it makes sense to have two credentials from the same authenticator in one group...

Evidlo:
Closes #311 

This looks good so far.  I'll ask to move the contents of `multifactor_auth.rst` into a new section in the README as that's where our examples currently live.

I'll hold off on merging until the [thread over at keepassxc](https://github.com/keepassxreboot/keepassxc/pull/10311) is mostly resolved.

BryanJacobs:
Sorry, accidentally auto-closed as I updated my branch. I will reopen this momentarily with your requested changes.

PR #379, opened 2024-03-03
Implement framework for flexible 2FA

BryanJacobs:
This adds support for using the hmac-secret FIDO extension to contribute keying material for a KeePass 4 file.

It does this by storing an additional XML statekeeping blob in the outer ("public") header. This blob is designed to hold a variety of different authentication factors, such as passwords, key files, and Yubikey challenge-response devices.

This is a reopening of the accidentally-closed #373 . Sorry about that.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -475,6 +475,57 @@ TOTP URI which can be passed to an OTP library to generate codes
    >>> pyotp.parse_uri(e.otp).now()
    799270
 
+multifactor authentication
+--------------------------
+
+PyKeePass supports securing a database using an arbitrary combination of "authentication factors".
+A single factor could be something like a password, a file, or a hardware authenticator.
+
+Factors are arranged into "factor groups". In order to open the database, *one* factor from
+each group must be provided.
+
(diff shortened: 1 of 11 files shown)


pykeepass/kdbx_parsing/common.py
```diff
@@ -188,6 +195,34 @@ def compute_master(context):
...
+                        next_byte=0xFF,
+                    )
+            }
+        )
+    else:
+        # Removing header entirely
+        if "public_custom_data" in kdbx.header.value.dynamic_header:
+            del kdbx.header.value.dynamic_header["public_custom_data"]
+
+    # Beyond Python 3.7, construct makes the base class of a Container be `dict` instead of `OrderedDict`
```
BryanJacobs, line 219: This is very unfortunate: the `construct` library dynamically changes which methods are available on a `Container` based on the version the Python interpreter reports.

If you run Python 3.5, you get all the `OrderedDict` methods. If you run Python 3.11 you get only `dict` methods.

Issue #376, opened 2024-02-26
All fields of items created in GNOME Secrets are hidden


julianfairfax:
See https://github.com/aivanovski/keepassvault/issues/208. All fields of items created in GNOME Secrets are shown as hidden in apps such as [KeePassVault](https://github.com/aivanovski/keepassvault]. Here is a test file: [Test.zip](https://github.com/aivanovski/keepassvault/files/14411001/Test.zip). cc @aivanovski @A6GibKm

A6GibKm:
Could you please create an entry both in Secrets and another app (e.g. keepassxc) with the same fields and same values and dump the contents (in xml format) of both?

Evidlo:
Should this be moved to the Secrets issues?  `ProtectInMemory` in not something that is set in PyKeePass.

A6GibKm:
Its not something I set either afaik, how would I even begin doing it?

A6GibKm:
> > What suggests it **should** look like that? I couldn't find the docs for this protectmemory attribute.
> 
> The fact that this attribute determines if a field is hidden or not, and only the password field should be hidden.

Where does it say that it should be hidden? Without extra information , this might be a stylistic choice made on the other apps.

A6GibKm:
We dont use any internals of pykeepass, @Evidlo is there any public api to set this?

Evidlo:
There's no API for this attribute.  I haven't seen `ProtectInMemory` before, only `Protected`.

Here's what a newly created entry looks like.

``` python
>>> from pykeepass import PyKeePass; kp = PyKeePass('test4.kdbx', 'password', 'test4.key')

>>> g = kp.groups[3]

>>> kp.add_entry(g, 'banana_entry', 'user', 'pass')
Entry: "foobar_group/subgroup/subgroup2/banana_entry (user)"

>>> kp.dump_xml('/tmp/dump.xml')
```

```
            <Entry>
              <UUID>KjVPGNbWEe6C170xALbYkw==</UUID>
              <Times>
                <CreationTime>hSxy3Q4AAAA=</CreationTime>
                <LastModificationTime>hSxy3Q4AAAA=</LastModificationTime>
                <LastAccessTime>hSxy3Q4AAAA=</LastAccessTime>
                <ExpiryTime>hSxy3Q4AAAA=</ExpiryTime>
                <Expires>False</Expires>
                <UsageCount>0</UsageCount>
                <LocationChanged>hSxy3Q4AAAA=</LocationChanged>
[...]

PR #381, opened 2024-03-06
entry: Do not set protected=True on setters

A6GibKm:
Setting a property would call `_set_string_field` which by default would set the protected status of the attribute as `"True"`.

In order to verify in tests if a property is protected we add a private method `_is_property_protected` which shares the code with `is_custom_property_protected`.

Fixes: https://github.com/libkeepass/pykeepass/issues/376

### DIFF
--- a/pykeepass/entry.py
+++ b/pykeepass/entry.py
@@ -98,9 +98,11 @@ class Entry(BaseElement):
         Args:
             key (str): name of field
             value (str): value of field
-            protected (bool): mark whether the field should be protected in memory
-                in other tools.  This property is ignored in PyKeePass and all
-                fields are decrypted immediately upon opening the database.
+            protected (bool or None): mark whether the field should be protected in memory
+                in other tools.  If None, value is either copied from existing field or field
+                is created with protected property unset.
+
+        Note: pykeepass does not support memory protection
         """
         field = self._xpath('String/Key[text()="{}"]/..'.format(key), history=True, first=True)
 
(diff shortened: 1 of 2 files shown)


A6GibKm:
**cc**: @julianfairfax

A6GibKm:
The newly introduced tests fails if one revers the commit.

Evidlo:
The test description doesn't match what I'm seeing in the test.  Shouldn't we set a property/field with something and check that the `Protected` attribute has not changed?

This would imply that the `_set_string_field(..., protected=None)`  leaves `Protected` as is.

I guess something like this on line 107:

    # leave Protected attribute alone if not specified
    protected = self._is_protected_property(field) if protected is None else protected

A6GibKm:
> The test description doesn't match what I'm seeing in the test. Shouldn't we set a property and check that the `Protected` attribute has not changed?
> 
> This would imply that the `_set_string_field(..., protected=None)` leaves `Protected` as is.
> 
> I guess something like this on line 107:
> 
> ```
> # copy Protected attribute from existing property
> protected = self._is_protected_property(field) if protected is None else protected
> ```

Yes, this prob would work best. But I have to say having a `Option<bool>` (in rust notation) feels weird in python.

Evidlo:
It boils down to whether the default behavior should be preserving attribute values when setting properties that already exist.

A6GibKm:
I implemented the tri state for protected, do note that we set the password and OTP to protected=True if the values were initialized by the library.

Issue #382, opened 2024-03-09
[4.0.7] regression -> gnome secrets crashes on expiration date check with missmatch of date format


fabiscafe:
* System: Arch Linux
* python 3.11.8
* pykeepass 4.0.7
* secrets 8.0

[Secrets](https://gitlab.gnome.org/World/secrets) is a password manager.

After importing https://github.com/libkeepass/pykeepass/pull/378 to make secrets even build and work (see the PR), secrets now crashes when trying to unlock the .kdbx. It looks like at least one of my expiration dates does have an incompatible format. This shouldn't happen inside a minor release. Rolling back to pykeepass 4.0.6 makes secrets work again.

This is the output I get when starting it from the terminal. 

```
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/gsecrets/unlock_database.py", line 212, in _unlock_callback
    database_manager.unlock_finish(result)
  File "/usr/lib/python3.11/site-packages/gsecrets/database_manager.py", line 124, in unlock_finish
    self.entries.splice(0, 0, [SafeEntry(self, e) for e in db.entries])
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/gsecrets/database_manager.py", line 124, in <listcomp>
    self.entries.splice(0, 0, [SafeEntry(self, e) for e in db.entries])
                               ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/gsecrets/safe_element.py", line 545, in __init__
    self._check_expiration()
  File "/usr/lib/python3.11/site-packages/gsecrets/safe_element.py", line 603, in _check_expiration
    if self.props.expired:
       ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/gi/_propertyhelper.py", line 394, in obj_get_property
    return getattr(self, name, None)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/gi/_propertyhelper.py", line 207, in __get__
    value = self.fget(instance)
            ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/gsecrets/safe_element.py", line 919, in expired
    return self.entry.expired
           ^^^^^^^^^^^^^^^^^^
[...]

PR #383, opened 2024-03-10
Use built-in isoformat support

City-busz:
The ISO formatted string can be stored in multiple formats. Use the built-in datetime.isoformat() and datetime.fromisoformat() functions, which able to handle all variants.

Fixes: https://github.com/libkeepass/pykeepass/issues/382


Evidlo:
Thanks for investigating this, but `.fromisoformat` on Python 3.6-3.10 does not support the 'Z' suffix to indicate timezone, so we need an additional `.replace('Z', '+00:00')` as mentioned in [this thread](https://bugs.python.org/issue35829)

City-busz:
Okay, I updated the MR to include this replace for compatibility.

PR #386, opened 2024-03-21
Return `None` if `deref` encouters a broken reference (fixes #384)

### LINKED ISSUE #384: crash if reference is broken
I think I found a bug in the following code:
https://github.com/libkeepass/pykeepass/blob/769ee25eca8a82943f59946aafe8c38246b99fe0/pykeepass/pykeepass.py#L704

This crashes when there is a broken reference in the `username` field.  

I messed with the references in my database and suddenly keepmenu crashed. I dug deeper and I think this is actually a bug of pykeepass. 

See the also the bug report in keepmenu:
https://github.com/firecat53/keepmenu/issues/160

br-olf:
Hi I tried to fix the issue #384 and wrote a small test but I didn't mess with the test databases so I used `add_entry` and `delete_entry` to set up my testcase.
If a reference is broken the code will now just return `None` and won't raise an exception. I think this is better and can be caught easily by the caller of `deref`.

### DIFF
--- a/pykeepass/pykeepass.py
+++ b/pykeepass/pykeepass.py
@@ -687,7 +687,7 @@ class PyKeePass:
             ref (str): KeePass reference string to another field
 
         Returns:
-            str or uuid.UUID
+            str, uuid.UUID or None if no match found
 
         [fieldref]: https://keepass.info/help/base/fieldrefs.html
         """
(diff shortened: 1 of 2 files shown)


A6GibKm:
Please give the PR and commit a more descriptive title.

Please also add to end of the commit description

```
Fixes: https://github.com/libkeepass/pykeepass/issues/384
```

br-olf:
Done :)

Evidlo:
Thanks for the fix.

PR #388, opened 2024-03-26
Fix composite key computation for BytesIO

janbrummer:
In case a keyfile as BytesIO has been read before, the next read will be empty. We need to ensure that we are reading the data from the beginning.

Add seek to start to fix it.

### DIFF
--- a/pykeepass/kdbx_parsing/common.py
+++ b/pykeepass/kdbx_parsing/common.py
@@ -1,13 +1,13 @@
 import base64
 import codecs
 import hashlib
+import io
 import logging
 import re
 import zlib
 from binascii import Error as BinasciiError
 from collections import OrderedDict
 from copy import deepcopy
-from io import BytesIO
 
 from construct import (
     Adapter,
@@ -128,6 +128,8 @@ def compute_key_composite(password=None, keyfile=None):
     # hash the keyfile
     if keyfile:
         if hasattr(keyfile, "read"):
+            if hasattr(keyfile, "seekable") and keyfile.seekable():
+                keyfile.seek(0)
             keyfile_bytes = keyfile.read()
         else:
             with open(keyfile, 'rb') as f:
@@ -194,7 +196,7 @@ class XML(Adapter):
 
     def _decode(self, data, con, path):
         parser = etree.XMLParser(remove_blank_text=True)
-        return etree.parse(BytesIO(data), parser)
+        return etree.parse(io.BytesIO(data), parser)
 
     def _encode(self, tree, con, path):
         return etree.tostring(tree)
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -875,6 +875,20 @@ class AttachmentTests3(KDBX3Tests):
 
 
 class PyKeePassTests3(KDBX3Tests):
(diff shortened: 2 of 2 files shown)


pykeepass/kdbx_parsing/common.py
```diff
@@ -128,6 +128,7 @@ def compute_key_composite(password=None, keyfile=None):
     # hash the keyfile
     if keyfile:
         if hasattr(keyfile, "read"):
+            keyfile.seek(0)
```
A6GibKm, line 131: This is a bit dangerous, you need to check

- That there is a seek method
- That [seekable](https://docs.python.org/3/library/io.html#io.IOBase.seekable)() returns True (and that the method exists)

see https://docs.python.org/3/library/io.html#class-hierarchy

A6GibKm:
Please add a test to showcase how this can corrupt a db and that it does not happen with this patch.

janbrummer:
> Please add a test to showcase how this can corrupt a db and that it does not happen with this patch.

Could you take this one over? I do have some higher priority stuff on my stack and this blocks Secrets.

A6GibKm:
> > Please add a test to showcase how this can corrupt a db and that it does not happen with this patch.
> 
> Could you take this one over? I do have some higher priority stuff on my stack and this blocks Secrets.

Sure, thanks for looking into it!! Make sure I am able to push to this branch.

A6GibKm:
@Evidlo Hi, this is ready for review.  For context, this is corrupting databases on Secrets https://gitlab.gnome.org/World/secrets/-/issues/537.

pykeepass/kdbx_parsing/common.py
```diff
@@ -127,7 +127,13 @@ def compute_key_composite(password=None, keyfile=None):
         password_composite = b''
     # hash the keyfile
     if keyfile:
-        if hasattr(keyfile, "read"):
+        if (
+            isinstance(keyfile, io.BufferedIOBase)
+            or isinstance(keyfile, io.TextIOBase)
+            or isinstance(keyfile, io.RawIOBase)
```
Evidlo, line 133: Is this a general way to check if an object is seekable?  Could we possibly just check `hasattr('seekable')`

A6GibKm: I used https://stackoverflow.com/questions/1661262/check-if-object-is-file-like-in-python as a reference. I am not sure if thats the best way to look at it tough.

tests/tests.py
```diff
@@ -875,6 +875,19 @@ def test_fields(self):
 
 
 class PyKeePassTests3(KDBX3Tests):
+    def test_consecutives_saves_with_stream(self):
```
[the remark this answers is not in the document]
janbrummer: Added a link to this PR.

pykeepass/kdbx_parsing/common.py
janbrummer: Actually you can use hasattr here. So let's give it a try.

A6GibKm: While you can, from the link above I have to wonder if it is the most sensible thing

PR #389, opened 2024-04-04
Add database name and description property

janbrummer:


janbrummer:
@Evidlo Could you please review this one?

janbrummer:
https://gitlab.gnome.org/World/secrets/-/merge_requests/878

janbrummer:
@A6GibKm Time for a review?

pykeepass/pykeepass.py
```diff
@@ -268,6 +268,42 @@ def entries(self):
         excluding history"""
         return self.find_entries()
 
+    @property
+    def database_name(self):
+        """Name of database"""
+        elem = self._xpath('/KeePassFile/Meta/DatabaseName', first=True)
```
A6GibKm, line 274: I see that in the other cases the `path` is inlined in the `_xpath` fn I would do them all the same, and if there is already a code style in this file, follow it.

pykeepass/pykeepass.py
```diff
@@ -268,6 +268,42 @@ def entries(self):
...
+
+    @database_name.setter
+    def database_name(self, name):
+        path = '/KeePassFile/Meta/DatabaseName'
+        item = self._xpath(path, first=True)
+        item.text = str(name)
+
+    @property
+    def database_description(self):
+        """Description of database"""
```
A6GibKm, line 285: of *the* database

pykeepass/pykeepass.py
```diff
@@ -268,6 +268,42 @@ def entries(self):
...
+    def database_description(self, name):
+        path = '/KeePassFile/Meta/DatabaseDescription'
+        item = self._xpath(path, first=True)
+        item.text = str(name)
+
+    @property
+    def default_username(self):
+        """Default Username"""
+        elem = self._xpath('/KeePassFile/Meta/DefaultUserName', first=True)
+        return elem.text
```
A6GibKm, line 299: can this return None? or will it return `""`? This should be documented.

A6GibKm (commented):
Thanks. Please add a test for the setters and getters. 

- get value
- set value to X
- get value and check that it matches X
- save db
- close db
- open db
- check the it is still X

A6GibKm:
Is this missing a rebase? I think python 3.6 tests were removed one or two weeks ago.

pykeepass/pykeepass.py
janbrummer: I've used recyclebin_group as base, so this is existing code style. What should i change then? There is no Coding Guidelines in this project..

pykeepass/pykeepass.py
janbrummer: Updated

pykeepass/pykeepass.py
janbrummer: In case it is not set it is None. Added description.

pykeepass/pykeepass.py
A6GibKm: If the file is consistent in its style follow that, if not choose one style of your liking and use the same for all 4 fns.

janbrummer: Done

janbrummer:
> Thanks. Please add a test for the setters and getters.
> 
>     * get value
> 
>     * set value to X
> 
>     * get value and check that it matches X
> 
>     * save db
> 
>     * close db
> 
>     * open db
> 
>     * check the it is still X

Added tests

janbrummer:
@Evidlo Can you please merge it?

Issue #390, opened 2024-04-04
Runtime dependency on pkg_resources module (from setuptools) is missing from pyproject.toml


stephen-dexda:
See https://github.com/libkeepass/pykeepass/blob/master/pykeepass/version.py

tandav:
Same here - library not working:

```
pip install pykeepass
python
>>> from pykeepass import PyKeePass
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/tandav/.cache/virtualenvs/tmp-fa63e4b506676ee/lib/python3.12/site-packages/pykeepass/__init__.py", line 4, in <module>
    from .version import __version__
  File "/Users/tandav/.cache/virtualenvs/tmp-fa63e4b506676ee/lib/python3.12/site-packages/pykeepass/version.py", line 4, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

Evidlo:
Fixed in master.  To be released in 4.1.0

PR #392, opened 2024-05-06
Package metadata version retrieval fix.

guillermodotn:
The package relies on `setuptools` to retrieve the package metadata version. However, this dependency is not included in `pyproject.toml`, leading to runtime errors, as described in issue #390.

Following the recommendation in the [`version.py`](https://github.com/libkeepass/pykeepass/blob/a2da6854af4c45487f6788fd693d921c26ca8eb5/pykeepass/version.py) module, the package will omit `pkg_resources.get_distribution` and instead utilize the built-in `importlib.metadata.version` method to retrieve the package version for Python version 3.8 and above.

For older Python versions (< 3.8), `setuptools` will be installed as a dependency.

### DIFF
--- a/pykeepass/version.py
+++ b/pykeepass/version.py
@@ -1,5 +1,10 @@
 __all__= ["__version__"]
 
-# FIXME: switch to using importlib.metadata when dropping Python<=3.7
-import pkg_resources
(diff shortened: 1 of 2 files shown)


guillermodotn:
Fixes: #390

PR #418, opened 2025-05-15
Allow password to be skipped or set to None in add_entry

rrasmuseklund:
Fix: Allow skipping password or `password=None` in `add_entry`

This PR fixes an issue where skipping or passing `None` as the `password` to `add_entry` would raise a `TypeError`, due to `lxml` not accepting `None` when building XML elements.

According to the documentation:
https://libkeepass.github.io/pykeepass/pykeepass.html#PyKeePass.add_entry

> password (str or None): the password of the new entry

The documentation already states that `None` is allowed, but the code did not support it.

### What this PR does

- Skips adding the `<Password>` field entirely if `password` is `None`
- Matches the KeePass GUI behavior when no password is provided

### DIFF
--- a/pykeepass/entry.py
+++ b/pykeepass/entry.py
@@ -41,9 +41,11 @@ class Entry(BaseElement):
             )
             self._element.append(E.String(E.Key('Title'), E.Value(title)))
             self._element.append(E.String(E.Key('UserName'), E.Value(username)))
-            self._element.append(
-                E.String(E.Key('Password'), E.Value(password, Protected="True"))
-            )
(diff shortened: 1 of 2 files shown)


spaetz:
I just proposed a similar solution in https://github.com/libkeepass/pykeepass/pull/430, also including title, username AND password as allowing None. Instead of skipping the entries if they are None, I set them to an empty string, so that the XML contains these entries. (I am not sure if they are optional).

Evidlo:
Sorry it took me so long to get around to this.  This was fixed in a later PR.

PR #429, opened 2026-03-09
pykeepass: Fix docs, remove unneeded indentation

### LINKED ISSUE #419: add_entry should return "Entry" not "Group"
Hello

In tthe file `pykeepass.py` starting at line 650, the `add_entry()` definition defines its return value as `Group`.  The ac tual return type is `Entry`.

I am new enough to git that I am not really sure how to submit the suggested fix.

spaetz:
add_entry returns an Entry, not a group. Fix the docs. Also, given the `if` path only ever raises an Exception, there is no need to put the rest of the code behind an `else` conditional. Less indendation is good!

Fixes #419

### DIFF
--- a/pykeepass/pykeepass.py
+++ b/pykeepass/pykeepass.py
@@ -673,7 +673,7 @@ class PyKeePass:
 
 
         Returns:
-            `Group`: newly added group
+            `Entry`: newly added entry
         """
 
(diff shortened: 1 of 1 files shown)


Evidlo:
Thanks for the contribution!

